PR #9, opened 2019-08-06
get output path from anywhere, and add predictor

jeremyasapp:
Implements a new way of getting the output path of a trial via a simple function
`get_trial_dir`


williamabrwolf:
@jeremyasapp @nmatthews-asapp ETA for shipping this PR? #bae

PR #16, opened 2019-08-07
better tutorial

jeremyasapp:


docs/tutorials/custom.rst
```diff
@@ -6,58 +6,111 @@ While Flambé offers a large number of `Component` objects to use in experiments
...
-          pipy. The latter allows you to specify a specific version of your extension. For github,
+          pypi. The latter allows you to specify a specific version of your extension. For github,
           we also support links to specific commit or branches.
+
+Flambé will require your extension to be installed. You can do so manually by running:
+
+``pip install my_package`` 
+
+or Flambé can install all the extensions specified in your
+configuration automatically if the ``-i`` flag is used:
```
nmatthews-asapp: Perhaps we say: "...automatically _while executing your config_ if the..."
to make it super clear the `-i` flag is just adding something to normal execution and we'll still be executing the config.

docs/tutorials/custom.rst
```diff
@@ -6,58 +6,111 @@ While Flambé offers a large number of `Component` objects to use in experiments
...
+
+Flambé configurations support any python object that inherits from ``flambe.Component``.
+You can decide to inherit from one of our base classes such as ``flambe.nn.Module`` or ``flambe.dataset.Dataset``,
+or you can inherit from ``flambe.Component`` directly.
+
+A ``Component`` must simply implement a ``run`` method which returns a boolean indicating whether execution
+should continue or not (useful for multi-step components such as a ``Trainer``).
+
+Additionally, if you would like to run hyperparameter search on your custom
+component, you must implement the ``metric`` method which returns the current best metric.
```
nmatthews-asapp: For the first time something is mentioned in a section we should link instead of just formatting, i.e. use `` :meth:`~flambe.compile.Component.metric` ``

Note that the `~` character means only the last item, in this case `metric`, will be shown

docs/tutorials/custom.rst
jeremyasapp: Will add that!

PR #24, opened 2019-08-09
fix version

jeremyasapp:
Fixing the pytorch version as 1.2.0 seems to break serialization


jeremyasapp:
Was done already in another PR..

PR #32, opened 2019-08-15
Hot fix in tmux checks

iitzco-asapp:
There was a bug in this check.

The idea is that sometimes `tmux` gives a timeout and the command fails. A new check is added to see if the service launched using `tmux` is running, prior to launching an error.


jeremyasapp:
Approved as a temporary fix. but we'll need to look into this more in the near future.

PR #35, opened 2019-08-16
Fix timeout feature!

iitzco-asapp:
The problem was that `ray` was still running in all instances after the remote experiment was over and this caused the AWS timeouts to not trigger (because ray caused CPU usage to go up).

Now when a remote experiment is over, flambe will call `ray stop` on all instances.

### DIFF
--- a/flambe/cluster/aws.py
+++ b/flambe/cluster/aws.py
@@ -5,6 +5,7 @@
 import boto3
 import botocore
 import logging
+from datetime import datetime, timedelta
 
 from typing import Generator, Dict, Tuple, List, TypeVar, Type, Any, Optional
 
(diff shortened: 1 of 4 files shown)


jeremyasapp:
Did you test whether this resolves the issue?

iitzco-asapp:
Also, in addition to the ray cluster shutdown, when using `AWSCluster` now changing timeouts in the same cluster is robust and consistent.

The problem was when:

1) You create a cluster with timeout `-1`.
2) It goes idle for 1 week.
3) You want to use it again but with timeout `5`.

When you execute the cluster in (3), then the alarm will be created and it will trigger right away (because the condition is met). So the instances will terminate before giving you the option of launching an experiment...

Now, before creating an alarm we add some fake CPU usage to fool the alarm and so that it doesn't consider past metrics.

PR #38, opened 2019-08-20
label frequencies

jeremyasapp:
Adds label frequencies, inverse frequencies and counts as tensors to be used in loss functions or elsewhere.

### DIFF
--- a/flambe/field/label.py
+++ b/flambe/field/label.py
@@ -1,13 +1,14 @@
-from typing import Optional, Dict
+from typing import Optional, Dict, Sequence
+from collections import OrderedDict as odict
 
 import torch
 import numpy as np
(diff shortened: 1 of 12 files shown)


flambe/field/label.py
```diff
@@ -1,37 +1,76 @@
...
 
         Parameters
         ----------
         one_hot : bool, optional
             Set for one-hot encoded outputs, defaults to False
         multilabel_sep : str, optional
-            If given, splits the input label into multiple labels using
-            the given separator, defaults to None.
+            If given, splits the input label into multiple labels
+            using the given separator, defaults to None.
```
iitzco-asapp: Missing docs for labels.

PR #39, opened 2019-08-21
Transformer jeremyasapp

jeremyasapp:
- [x] Base Transformer Implementation based on Pytorch's, with a couple things changed to improve the interfaces
- [x] Added an SRU option to replace the FFN
- [x] Positional Embeddings 

Breaking changes:
- nn.embedder becomes nn.embedding
- mask in RNNEncoder becomes padding_mask to match Transformer
- BaseSampler now has batch_first=False by default for sequential data


flambe/nn/transformer.py
```diff
@@ -0,0 +1,532 @@
...
+        use_sru: bool, optional
+            If true, replaces the FFN with an SRU.
+
+        """
+        super(TransformerEncoderLayer, self).__init__()
+
+        self.self_attn = MultiheadAttention(d_model, nhead, dropout=dropout)
+
+        if use_sru:
+            self.sru = SRU(d_model, dim_feedforward, num_layers=1, dropout=dropout)
```
taoleicn: use `SRUCell` to avoid extra computation overhead.

flambe/nn/transformer.py
```diff
@@ -0,0 +1,532 @@
...
+class TransformerEncoder(Module):
+    """TransformerEncoder is a stack of N encoder layers."""
+
+    def __init__(self,
+                 d_model: int,
+                 nhead: int,
+                 num_layers: int,
+                 dim_feedforward: int = 2048,
+                 dropout: float = 0.1,
+                 use_sru: bool = False) -> None:
```
taoleicn: A different option is to provide `custom_forward_encoder: Module = None` as a flexible choice to replace FFN.
For example, depending on the application, the SRUCell can be uni-directional or bi-directional. Or we may want to explore a different hyper-parameter setup for SRUCell. This avoids adding SRU args to the __init__ method?

You can create  `nn.TransformerSRU` which then creates SRUCell() for each layer and pass that as the custom_forward_encoder?

flambe/nn/transformer.py
taoleicn:
```
SRUCell(d_model, dim_feedforward, 
        dropout=dropout, 
        has_skip_term=False, 
        is_input_normalized=True)
```

taoleicn: `bidirectional = True` or `False` should be configurable.

flambe/nn/transformer.py
```diff
@@ -0,0 +1,532 @@
...
+        src_key_padding_mask: torch.Tensor, optional
+            The mask for the src keys per batch (optional).
+
+        """
+        src2 = self.self_attn(src, src, src, attn_mask=src_mask,
+                              key_padding_mask=src_key_padding_mask)[0]
+        src = src + self.dropout1(src2)
+        src = self.norm1(src)
+        if self.use_sru:
+            src2 = self.linear2(self.sru(src))
```
taoleicn: specify padding mask for SRU:
`self.sru(src, mask_pad=src_key_padding_mask)`

flambe/nn/transformer.py
```diff
@@ -0,0 +1,532 @@
...
+
+        """
+        super(TransformerDecoderLayer, self).__init__()
+
+        self.use_sru = use_sru
+        self.self_attn = MultiheadAttention(d_model, nhead, dropout=dropout)
+        self.multihead_attn = MultiheadAttention(d_model, nhead, dropout=dropout)
+
+        if use_sru:
+            self.sru = SRU(d_model, dim_feedforward, num_layers=1, dropout=dropout)
```
taoleicn: see my comments for `TransformerEncoder`

flambe/nn/transformer.py
```diff
@@ -0,0 +1,532 @@
...
+        tgt2 = self.self_attn(tgt, tgt, tgt, attn_mask=tgt_mask,
+                              key_padding_mask=tgt_key_padding_mask)[0]
+        tgt = tgt + self.dropout1(tgt2)
+        tgt = self.norm1(tgt)
+        tgt2 = self.multihead_attn(tgt, memory, memory, attn_mask=memory_mask,
+                                   key_padding_mask=memory_key_padding_mask)[0]
+        tgt = tgt + self.dropout2(tgt2)
+        tgt = self.norm2(tgt)
+        if self.use_sru:
+            tgt2 = self.linear2(self.sru(tgt))
```
taoleicn: specify padding mask for SRU:
`self.sru(tgt, mask_pad=tgt_key_padding_mask)`

flambe/nn/transformer.py
taoleicn: TransformerSRU may also support you to pass the hidden states of the last data segment (chunk) as the init hidden states of the current data segment. 

This is useful for training corpus language model with BPTT.

flambe/nn/__init__.py
```diff
@@ -1,13 +1,15 @@
 from flambe.nn.module import Module
 from flambe.nn.softmax import SoftmaxLayer
 from flambe.nn.mos import MixtureOfSoftmax
-from flambe.nn.embedder import Embeddings, Embedder
+from flambe.nn.embedding import Embeddings, Embedder
```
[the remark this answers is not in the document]
jeremyasapp: People usually just import from `flambe.nn` so I don't think this would be a big deal. It's just a much better name now that there is extra stuff in this module.

flambe/nn/transformer.py
```diff
@@ -0,0 +1,532 @@
...
+    def __init__(self,
+                 d_model: int = 512,
+                 nhead: int = 8,
+                 num_encoder_layers: int = 6,
+                 num_decoder_layers: int = 6,
+                 dim_feedforward: int = 2048,
+                 dropout: float = 0.1,
+                 use_sru: bool = False,
+                 custom_encoder: Optional[Module] = None,
+                 custom_decoder: Optional[Module] = None) -> None:
```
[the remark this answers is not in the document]
jeremyasapp: Yeah it's part of the original implementation, presumably so you can do things like adding embeddings on top? I think it kind of has to do with what is getting passed to what object in the forward, which would be akward to generalize

flambe/nn/transformer.py
jeremyasapp: Yeah so the question here is that if I make a TransformerSRU, I also need to make a TransformerSRUEncoder, TransformerSRUDecoder, etc.. Seems like a lot of repeated code thats hard to avoid.

I considered the idea of passing a custom_forward_module. We could definitely do that, or take additional **kwargs to be passed to SRU. I think the custom_forward_module idea makes the most sense so that you can pass in anything you want, doesn't even have to be SRU. Thoughts?

jeremyasapp: I like the idea of custom_forward_module and very simple TransformerSRU.

Regarding the state, maybe i should add that to the main implementation, and the feedforward doesn't use it?

flambe/nn/transformer.py
jeremyasapp: Cool, will address this as part of your TransformerSRU suggestion

flambe/nn/transformer.py
jeremyasapp: great catch

flambe/nn/transformer.py
taoleicn: > Regarding the state, maybe i should add that to the main implementation, and the feedforward doesn't use it?
you mean add it as an optional value and default to None? sounds good to me.

flambe/nn/transformer.py
taoleicn: i believe it should be `src_key_padding_mask` instead of `1 - src_key_padding_mask` based on the doc string. and it should be Byte tensor already.

flambe/nn/transformer_sru.py
```diff
@@ -0,0 +1,307 @@
...
+        """
+        tgt2 = self.self_attn(tgt, tgt, tgt, attn_mask=tgt_mask,
+                              key_padding_mask=tgt_key_padding_mask)[0]
+        tgt = tgt + self.dropout1(tgt2)
+        tgt = self.norm1(tgt)
+        tgt2 = self.multihead_attn(tgt, memory, memory, attn_mask=memory_mask,
+                                   key_padding_mask=memory_key_padding_mask)[0]
+        tgt = tgt + self.dropout2(tgt2)
+        tgt = self.norm2(tgt)
+        tgt2 = self.linear2(self.sru(tgt, state=memory, mask_pad=tgt_key_padding_mask))
```
taoleicn: `state` is not an argument for SRUcell. (should have renamed `c0` into `state` in the future for SRU)
for now you can do `self.sru(tgt, memory, mask_pad=..)` instead

flambe/nn/transformer_sru.py
```diff
@@ -0,0 +1,307 @@
...
+            The mask for the src sequence (optional).
+        src_key_padding_mask: torch.Tensor, optional
+            The mask for the src keys per batch (optional).
+
+        """
+        src2 = self.self_attn(src, src, src, attn_mask=src_mask,
+                              key_padding_mask=src_key_padding_mask)[0]
+        src = src + self.dropout1(src2)
+        src = self.norm1(src)
+        src2 = self.linear2(self.sru(src, state=memory, mask_pad=src_key_padding_mask))
```
taoleicn: state is not an argument for SRUcell. (should have renamed c0 into state in the future for SRU)
for now you can do self.sru(tgt, memory, mask_pad=..) instead

flambe/nn/transformer_sru.py
```diff
@@ -0,0 +1,307 @@
...
+
+        Extra keyword arguments are passed to the SRUCell.
+
+        """
+        super(TransformerSRUEncoderLayer, self).__init__()
+
+        self.self_attn = nn.MultiheadAttention(d_model, nhead, dropout=dropout)
+        self.sru = SRUCell(d_model,
+                           dim_feedforward,
+                           dropout,
```
taoleicn: oh i forgot one thing. maybe introduce a `sru_dropout` here option for compatibility with the machine translation experiments of the SRU paper. i mean sth:
```
self.sru = SRUCell(d_model,
        dim_feedforward,
        dropout if sru_dropout == -1 else sru_dropout,
        ...)
```

jeremyasapp: Done

jeremyasapp:
> Just want to do a sanity check: do any of these classes need custom state / register_attrs calls?

They do not

flambe/nn/embedding.py
```diff
@@ -0,0 +1,306 @@
...
+        padding_idx: int, optional
+            Passed the nn.Embedding object. See pytorch documentation.
+
+        """
+        super().__init__()
+
+        self.embedding = embedding
+        self.dropout = nn.Dropout(embedding_dropout)
+        self.encoder = encoder
+        self.pooling = pooling
```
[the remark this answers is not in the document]
jeremyasapp: what kind of instance check would we even do? It can be any Module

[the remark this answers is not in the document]
jeremyasapp: Yeah I don't think that makes sense to discuss in this PR, this is something that we would need to do everywhere if we decided to do so

PR #40, opened 2019-08-21
Ignore gitignore when rsyncing flambe

iitzco-asapp:
- In dev mode, when rsyncing flambe to the instances all content in gitignore will be ignored. **This makes launching new clusters significantly faster**

- Also, I added a fix in the gitignore file that was not allowing `bin/` to be included.


jeremyasapp:
Awesome

PR #45, opened 2019-08-22
Edit .gitignore to fix rsync in dev mode

iitzco-asapp:
This was not allowing sending the `instance` code to the instances when running in `dev` mode, so it needs to be removed from the gitignore.


iitzco-asapp:
No, we are always executing `rsync` pointing to the folder where `flambe` repo is located, so not related to `cwd`.

I think this is related to the fact that `rsync` is not being consistent with the usage of our `.gitignore` file. When it sees `instance/`, it ignores any folder that matches the name, not just a top level folder as git would do.

jeremyasapp:
In that case, I wonder if it would be beneficial to add a test that ensures that no folder inside flambe have a name that would be ignored by gitignore, so that this doesn't happen in the future. For example, it looks like the name `site` and `target` would be prohibited.

PR #47, opened 2019-08-22
New, Explicit Links

### LINKED ISSUE #46: Allow more fine-grained control of links in YAML configs
**Is your feature request related to a problem? Please describe.**
Currently it's not very explicit how links are resolved i.e. are they resolved against the config (nested dictionary structure) or against the attributes of the objects once they are initialized. The answer is both; it depends on what you're linking to in the config and where that link is.

**Describe the solution you'd like**
Change links to be most more powerful and more intuitive by supporting pointing to a specific object in the config and then accessing attributes on that object. This will look like:
```
!@ train_stage[model][embedder][encoder].rnn.hidden_size
```
for example. The brackets work similar to brackets in normal python, they access a specific object in the nested dictionary structure created from the YAML config. The dot notation is then used to access attributes on the initialized object.

**Describe alternatives you've considered**
We've considered leaving it implicit, or dropping the attribute access altogether, but these limitations are confusing and, well, limiting. We also considered other syntax for exactly the same logic, but this familiar syntax should be intuitive to new users as it mirrors how things are done in Python.

nmatthews-asapp:
Resolves #46 by implementing the suggested syntax for links e.g.
```
training_stage[model][embedder][encoder].rnn.hidden_size
```
Where the bracket notation is used to navigate the nested dictionary structure of the config, and the dot notation is used to access attributes on the initialized object, reflecting python syntax more closely, and making the distinction between the two explicit


flambe/compile/serialization.py
```diff
@@ -165,10 +165,11 @@ def _update_link_refs(schema: Mapping) -> None:
     from flambe.compile.component import Link
     for _, _, value in traverse(schema):
         if isinstance(value, Link):
-            value.obj = schema
+            # value.obj = schema
```
jeremyasapp: commented code

flambe/compile/serialization.py
```diff
@@ -165,10 +165,11 @@ def _update_link_refs(schema: Mapping) -> None:
...
+            value.target = schema[value.root_schema]
             # TODO temporary hack until pipeline + load linking are
             # unified
-            value.attr = [value.var_name] + value.attr
+            # value.attr = [value.var_name] + value.attr
```
jeremyasapp: commented code

flambe/experiment/utils.py
```diff
@@ -92,7 +92,7 @@ def check_search(blocks: Dict[str, Schema],
             raise SearchComponentError(block_id)
 
 
-def convert_tune(blocks: Mapping[str, Schema]):
+def convert_tune(data: Mapping[str, Schema]):
     """Convert the options and links in the block.
 
     Convert Option objects to tune.grid_search or
```
jeremyasapp: this docstring seems outdated with respect to the inputs

flambe/experiment/utils.py
```diff
@@ -369,14 +373,14 @@ def extract_needed_blocks(schemas: Dict[str, Schema],
 
     # Get this block's links
     for _, _, value in traverse(this_block):
-        if isinstance(value, Link) and value.var_name != block_id:
+        if isinstance(value, Link) and value.root_schema != block_id:
             # TODO Add new block_id attribute to Link
             # Ensure intra-block links are not added to prevent inf loop
-            needed.add(value.var_name)
+            needed.add(value.root_schema)
         elif isinstance(value, Iterable):
```
jeremyasapp: strings are iterable is that ok?

flambe/field/label.py
```diff
@@ -37,16 +37,16 @@ def __init__(self,
         """
         self.one_hot = one_hot
         self.multilabel_sep = multilabel_sep
-        self.tokenizer = LabelTokenizer(self.multilabel_sep)
+        self.tokenizer = LabelTokenizer(multilabel_sep=self.multilabel_sep)
```
jeremyasapp: why do we need to add the keyword arg here?

flambe/nn/rnn.py
```diff
@@ -196,15 +196,15 @@ def __init__(self,
...
-        self.rnn = RNNEncoder(input_size,
-                              hidden_size,
-                              n_layers,
-                              rnn_type,
-                              dropout,
-                              bidirectional,
-                              layer_norm,
-                              highway_bias,
-                              rescale)
+        self.rnn = RNNEncoder(input_size=input_size,
```
jeremyasapp: Is this a new requirement? What happens if we don't

flambe/experiment/utils.py
nmatthews-asapp: it is ok here

PR #76, opened 2019-09-11
Improve console streaming

iitzco-asapp:
Now the console is more robust by removing the streaming strategy.


iitzco-asapp:
Instead, see #78

PR #79, opened 2019-09-13
make everything batch first

jeremyasapp:
This PR makes Flambé fully batch first, for better generalizability and addresses a bug in the BaseSampler when all elements in the batch are the same size.

### DIFF
--- a/flambe/sampler/episodic.py
+++ b/flambe/sampler/episodic.py
@@ -8,7 +8,11 @@ from flambe.sampler import Sampler
 
 
 class EpisodicSampler(Sampler):
-    """Implements an EpisodicSample object."""
+    """Implement an EpisodicSample object.
+
+    Currently only supports sequence inputs.
+
+    """
 
     def __init__(self,
                  n_support: int,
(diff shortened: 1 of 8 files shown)


flambe/sampler/episodic.py
```diff
@@ -113,12 +112,12 @@ def sample(self,
                 support_source, support_target = list(zip(*supports))
 
                 query_source = pad_sequence(query_source,
-                                            batch_first=self.batch_first,
+                                            batch_first=False,
```
iitzco-asapp: this should be `True`

flambe/sampler/episodic.py
```diff
@@ -113,12 +112,12 @@ def sample(self,
...
                                             padding_value=self.pad)
                 query_target = torch.tensor(query_target)
 
                 support_source = pad_sequence(support_source,
-                                              batch_first=self.batch_first,
+                                              batch_first=False,
```
iitzco-asapp: This also should be `True`

jeremyasapp: lol

PR #80, opened 2019-09-13
Fix last and first pooling

iitzco-asapp:
Fixes pooling according to the batch first approach.


iitzco-asapp:
Tests are needed before merging.

PR #81, opened 2019-09-16
Add pooling tests

iitzco-asapp:
Adds unit testing for all pooling options.


iitzco-asapp:
If this is merged, we need to run some tests to see how much overhead this adds to the forward pass.

iitzco-asapp:
This is on hold because of the possible overhead performance issues this implementation could have.

flambe/nn/pooling.py
```diff
@@ -83,7 +83,8 @@ def forward(self,
             The output data, as a tensor of shape [B x H]
 
         """
-        padding_mask = padding_mask or _default_padding_mask(data)
+        if padding_mask is None:
+            padding_mask = _default_padding_mask(data)
```
iitzco-asapp: boolean operations over tensors are not recommended.

PR #94, opened 2019-10-01
Update requirements.txt

williamabrwolf:
Does this have to be here?


jeremyasapp:
No this is not needed. We have a requirments-gpu.txt which contains sru[cuda] and is installed with flambe[cuda]

PR #100, opened 2019-10-01
Add config to the webapp

iitzco-asapp:


flambe/runnable/runnable.py
```diff
@@ -27,6 +27,19 @@ class Runnable(MappedRegistrable):
     def __init__(self, **kwargs) -> None:
         self.config = configparser.ConfigParser()
         self.extensions: Dict[str, str] = {}
+        self.content: Optional[str] = None
+
+    def inject_content(self, content: str) -> None:
```
jeremyasapp: in theory you should be able to just dump the experiment to YAML right?

iitzco-asapp: Yes, but that would add the default parameters. I want to display the exact config the user used.

PR #102, opened 2019-10-07
Auto publish CI

iitzco-asapp:
CircleCI will also handle publishing to PyPI.


iitzco-asapp:
I suggest merging this once we are able to test `end2end` job with larger containers... as the publishing job will require `end2end`.

PR #106, opened 2019-10-08
Add stop on fail flag to Experiment

nmatthews-asapp:
This PR adds a flag `stop_on_failure` to experiment that indicates whether a failed trial should terminate the entire experiment


flambe/experiment/experiment.py
```diff
@@ -342,10 +344,17 @@ def trial_name_creator(trial):
...
 
+                any_error = False
                 for t in trials:
                     if t.status == t.ERROR:
                         logger.error(f"{t} ended with ERROR status.")
                         success[block_id] = False
+                        any_error = True
+                if any_error and self.stop_on_failure:
+                    self.teardown()
+                    logger.error(f"Stopping experiment at block '{block_id}' because there was an "
```
iitzco-asapp: we could add colors here. `cl.RE` for red one.

iitzco-asapp: Check this line https://github.com/asappresearch/flambe/pull/106/files#diff-7450e432b0e35241531af760c8a6d4eaR410

for reference

iitzco-asapp:
Also I think we need a little more work on the progress state so that the report site shows this new state.

flambe/experiment/experiment.py
nmatthews-asapp: `self.teardown()` updates the status I think this is okay

jeremyasapp:
@iitzco-asapp do you want to add that in or should we go ahead and merge as is?

PR #110, opened 2019-10-16
Update reqs and other improvements

iitzco-asapp:
- Update `sru` version in requirements
- Add a new attribute in text field to track unknown numericals.

### DIFF
--- a/flambe/field/text.py
+++ b/flambe/field/text.py
@@ -79,8 +79,8 @@ class TextField(Field):
             given a random embedding from a normal distribution.
             Otherwise, all of them map to the '<unk>' token.
         drop_unknown: bool
-            Whether to drop tokens that don't have embeddings associated.
-            Defaults to True.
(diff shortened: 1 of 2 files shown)


flambe/field/text.py
```diff
@@ -169,13 +171,16 @@ def setup(self, *data: np.ndarray) -> None:
...
                             self.vocab[token] = index = index + 1
                             embeddings_matrix.append(torch.tensor(model[token]))
-                        elif self.unk_init_all:
-                            # Give every OOV it's own embedding
-                            self.vocab[token] = index = index + 1
-                            embeddings_matrix.append(torch.randn(model.vector_size))
                         else:
-                            # Collapse all OOV's to the same token id
-                            self.vocab[token] = self.vocab[self.unk]
+                            if self.unk_init_all:
```
jeremyasapp: What is going on here? Why did you indent this in? Now you may assign two ID's to the same token

Issue #103, opened 2019-10-08
Expose activation parameter for SRU


iitzco-asapp:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
When using `sru` as rnn type it could be useful to expose an activation parameter: 
https://github.asapp.dev/ASAPPinc/prodml/blob/062c2d10f0475acd9f2fca9659bebf1450951038/asapp/model/sru/sru_functional.py#L444

iitzco-asapp:
Solved in #114

PR #113, opened 2019-10-18
[OPTION 1] Expose the activation parameter

iitzco-asapp:
One option to solve https://github.com/asappresearch/flambe/issues/103


iitzco-asapp:
#114 was merged instead.

PR #115, opened 2019-10-21
New Pipeline Tokenizer

iitzco-asapp:
This PR proposes a new tokenizer, `PipelineTokenizer` that could be used as a base class when several sequential transformations should be applied to an input string.


jeremyasapp:
As discussed offline, this doesn't really add much at the moment, so closing for now.

PR #120, opened 2019-10-23
Add -s flag shortcut for secrets

iitzco-asapp:


flambe/runner/run.py
```diff
@@ -112,7 +112,7 @@ def main(args: argparse.Namespace) -> None:
     parser.add_argument('-f', '--force', action='store_true', default=False,
                         help='Override existing runnables. Be careful ' +
                              'when using this flag as it could have undesired effects.')
-    parser.add_argument('--secrets',
+    parser.add_argument('-s', '--secrets',
                         type=str, default=os.path.join(FLAMBE_GLOBAL_FOLDER, "secrets.ini"))
```
jeremyasapp: do we want this to be the default place? I kind like this: ~/.flambe.conf

iitzco-asapp: For the secrets lets keep it names `secrets` imo

PR #124, opened 2019-10-23
New schedulers

jeremyasapp:
- WarmupLinearScheduler
- NoamScheduler

### DIFF
--- a/flambe/learn/distillation.py
+++ b/flambe/learn/distillation.py
@@ -66,7 +66,8 @@ class DistillationTrainer(Trainer):
         scheduler : torch.optim.lr_scheduler._LRScheduler, optional
             An optional learning rate scheduler
         iter_scheduler : torch.optim.lr_scheduler._LRScheduler, optional
-            An optional learning rate scheduler to run after each ite
+            An optional learning rate scheduler to run after each batch
+            (i.e iteration)
         device: str, optional
             The device to use in the computation. Only used by compile.
         max_steps : int, optional
--- a/flambe/learn/distillation.py
+++ b/flambe/learn/distillation.py
@@ -32,6 +32,7 @@ class DistillationTrainer(Trainer):
                  metric_fn: Metric,
                  optimizer: Optimizer,
                  scheduler: Optional[_LRScheduler] = None,
+                 iter_scheduler: Optional[_LRScheduler] = None,
                  device: Optional[str] = None,
                  max_steps: int = 10,
                  epoch_per_step: float = 1.0,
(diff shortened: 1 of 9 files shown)


flambe/learn/distillation.py
```diff
@@ -64,6 +65,8 @@ def __init__(self,
             The optimizer to use
         scheduler : torch.optim.lr_scheduler._LRScheduler, optional
             An optional learning rate scheduler
+        iter_scheduler : torch.optim.lr_scheduler._LRScheduler, optional
+            An optional learning rate scheduler to run after each ite
```
iitzco-asapp: `iteration` typo

flambe/learn/train.py
```diff
@@ -219,14 +219,9 @@ def _train_step(self) -> None:
 
                 # Update iter scheduler
                 if self.iter_scheduler is not None:
-                    log(f'{tb_prefix}Training/LR', self.iter_scheduler.get_lr()[0], global_step)
-                    if isinstance(self.iter_scheduler, ReduceLROnPlateau):
-                        self.iter_scheduler.step(val_loss)
```
iitzco-asapp: Isn't this case valid anymore?

jeremyasapp: On each iteration you can't look at the val loss really. We could expand to provide a closure based on the training loss I guess? Maybe let's keep it like this for now until the need arises

flambe/nn/mlp.py
```diff
@@ -34,10 +34,11 @@ def __init__(self,
             Input_dimension
         output_size: int
             Output dimension
-        n_layers: int
+        n_layers: int, optional
             Number of layers in the network, defaults to 1
         dropout: float, optional
-            Dropout to be used before each MLP layer
+            Dropout to be used before each MLP layer.
+            Only used if n_layers > 1.
```
iitzco-asapp: Nice detail.

flambe/optim/linear.py
```diff
@@ -0,0 +1,53 @@
...
+            The number of linear warmup phases
+        n_steps : int, optional
+            The index of last step. Default: -1
+
+        """
+        self.warmup = warmup
+        self.n_steps = n_steps
+        super().__init__(optimizer, lr_lambda=self.lr_lambda, last_epoch=n_steps)  # type: ignore
+
+    def lr_lambda(self, step: int) -> float:
```
iitzco-asapp: Lets add unit tests for this method.

iitzco-asapp: I see some risk in touching this code without being careful

flambe/optim/noam.py
```diff
@@ -0,0 +1,56 @@
...
+            The number of linear warmup phases
+        d_model : int, optional
+            The index of last step. Default: -1
+
+        """
+        self.warmup = warmup
+        self.d_model = d_model
+        super().__init__(optimizer, lambda_lr=self.lr_lambda, last_epoch=-1)  # type: ignore
+
+    def lr_lambda(self, step: int) -> float:
```
iitzco-asapp: Same here, lets add some tests

flambe/learn/train.py
iitzco-asapp: Does this mean that we will not support `ReduceLROnPlateau`?

jeremyasapp: We do for the normal val step schedulers but it doesn't make sense for the iter scheduler

PR #129, opened 2019-10-29
Freeze reqs

iitzco-asapp:
Give upper bound for requirements assuming requirements use https://www.python.org/dev/peps/pep-0440/ and semantic versioning


iitzco-asapp:
There is a nice notation `~=` that we can use instead of repeating `>=X.Y.Z, ==X.Y.*` (https://www.python.org/dev/peps/pep-0440/#compatible-release)

PR #132, opened 2019-10-30
Fix links in lists in config

### LINKED ISSUE #130: Fix link reference resolution inside of lists
**Describe the bug**
Currently links inside of lists don't get their targets updated properly before compilation. This should be easily fixable by having the `traverse` method used in this utility function also recurse on sequences.

nmatthews-asapp:
Resolves #130 

This PR fixes a bug where links inside of lists in a YAML config are not updated to point to the right blocks, and consequently cannot be compiled.


tests/integration/examples/experiments/tc.yaml
```diff
@@ -30,3 +30,6 @@ pipeline:
       params: !@ b1[model].trainable_params
     max_steps: 2
     iter_per_step: 1
+    extra_validation_metrics:
```
iitzco-asapp: How does this example test this case? According to the description this is for lists of links right?

flambe/experiment/utils.py
```diff
@@ -145,6 +145,29 @@ def traverse(nested: Mapping[str, Any], path: Optional[List[str]] = None) -> Ite
...
+    ----------
+    obj: Any
+
+    Returns
+    -------
+    Iterable[Any]
+        Iterable of child values to obj
+
+    """
+    if isinstance(obj, Mapping):
```
jeremyasapp: I noticed that sets are not instances of Mapping or Sequence but they are instances of Iterable

PR #134, opened 2019-10-30
New resources mechanism

iitzco-asapp:
Resource are not anymore:
```
resources:
  local:
    emb: /path/to/local/embeddings
  remote:
    other_emb: /path/to/embedding/in/cluster
```

Instead:

```
resources:
    emb: /path/to/local/embeddings
    other_emb: !cluster /path/to/embedding/in/cluster
```

And even more!:

```
resources:
    emb: s3://bucket/embeddings
    other_emb: !cluster s3://bucket/embeddings
```

Using `!cluster` with remote URLs will leave the cluster to do the download.

### DIFF
--- a/flambe/compile/downloader.py
+++ b/flambe/compile/downloader.py
@@ -143,7 +143,7 @@ def download_s3_folder(url: str, destination: str) -> None:
 
 
 @contextmanager
-def download_manager(path: str):
+def download_manager(path: str, folder: str = None):
     """Manager for downloading remote URLs
 
     Parameters
@@ -152,6 +152,8 @@ def download_manager(path: str):
         The remote URL to download. Currently, only S3 and http/https
         URLs are supported.
         In case it's already a local path, it yields the same path.
+    folder: str
+        Optional folder where all remote content will be downloaded.
 
     Examples
     --------
@@ -168,6 +170,10 @@ def download_manager(path: str):
     """
     url = urlparse(path)
 
+    if not folder:
+        tmp_dir = tempfile.TemporaryDirectory()
+        folder = tmp_dir.name
+
     if not url.scheme:
         # 'path' is a local path
         if os.path.exists(os.path.expanduser(path)):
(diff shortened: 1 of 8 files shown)


flambe/compile/downloader.py
```diff
@@ -176,29 +183,33 @@ def download_manager(path: str):
             raise ValueError(f"Path: '{path}' does not exist locally.")
 
     else:
+        trailing_url = url.path[:-1] if url.path.endswith('/') else url.path
+        fname = trailing_url[trailing_url.rfind('/') + 1:]
```
jeremyasapp: Could this result in file collisions? Why not use the key in the resources dict for the file name?

flambe/experiment/options.py
```diff
@@ -131,3 +131,22 @@ def __repr__(self) -> str:
     @classmethod
     def to_yaml(cls, representer: Any, node: Any, tag: str) -> Any:
         return representer.represent_sequence('!g', node.elements)
+
+
+@alias('cluster')
```
jeremyasapp: I would recommend the name remote over cluster since we may some day send to a machine and not a cluster

flambe/compile/downloader.py
iitzco-asapp: Yes, def we can have collisions. Using the key in the resources is a good idea definitely.

flambe/experiment/options.py
iitzco-asapp: Ok! I don't have strong opinion about this naming.

iitzco-asapp: Done

flambe/compile/downloader.py
iitzco-asapp: Fixed.

flambe/experiment/options.py
iitzco-asapp: I'll bing this up again. When using a single remote machine, that will be also a "cluster" (of a single machine). That implementation will most likely follow our cluster paradigm (the orchestrator and factory will be the same machine, but logically if will be separated).

I think `cluster` is a better option here to keep consistency.

jeremyasapp: @nmatthews-asapp opinion?

iitzco-asapp: Discussed offline. `!cluster` for now!

PR #136, opened 2019-11-05
Issue 135 safer load

nmatthews-asapp:
This PR removes the behavior where flambe's `load` and `save` functions would use the current working directory for temporary files (mainly when working with compressed archives) and manually handle cleanup. We now use more standard and safe Python built-ins for working with temporary directories via `tempfile`


iitzco-asapp:
@nmatthews-asapp does this fix also the random folders being created when compressing?

PR #138, opened 2019-11-08
allow overwrite in save

### LINKED ISSUE #137: Checkpointing sometimes fails because of new safety check
**Describe the bug**
Because of a new safety check introduced in #136 checkpointing may fail when overwriting a previous save file. It *should* be possible to overwrite a file when saving. Whether or not it should be opt-in or opt-out is open for discussion.

nmatthews-asapp:
Resolves #137 
We need some way of allowing overwriting when appropriate. Not yet sure if this should be opt-in or opt-out.

### DIFF
--- a/flambe/compile/component.py
+++ b/flambe/compile/component.py
@@ -1307,7 +1307,7 @@ class Component(Registrable):
         return instance
 
     def save(self, path: str, **kwargs: Any):
-        flambe_save(self, path)
+        flambe_save(self, path, **kwargs)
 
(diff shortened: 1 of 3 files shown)


iitzco-asapp:
Nice catch. Is `override` more standard rather than `overwrite`?

nmatthews-asapp:
> Nice catch. Is `override` more standard rather than `overwrite`?

I think overwrite is more appropriate here, as the operation is "writing" a save file

PR #140, opened 2019-11-08
Replace gensim's deprecated fasttext method with currently supported one

siddharth-asapp:
gensim.models.KeyedVectors.load_fasttext_format does not seem to work for fasttext binary files. Instead gensim.models.fasttext.load_facebook_vectors seems to work however it by default accepts only .bin files.

### DIFF
--- a/flambe/field/text.py
+++ b/flambe/field/text.py
@@ -146,8 +146,8 @@ class TextField(Field):
                                                           binary=self.embeddings_binary)
             elif self.embeddings_format == 'fasttext':
                 model = fasttext.load_facebook_vectors(self.embeddings)
-                # model = KeyedVectors.load_fasttext_format(self.embeddings,
-                #                                          binary=self.embeddings_binary)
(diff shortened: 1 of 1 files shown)


flambe/field/text.py
```diff
@@ -144,8 +145,9 @@ def setup(self, *data: np.ndarray) -> None:
                 model = KeyedVectors.load_word2vec_format(self.embeddings,
                                                           binary=self.embeddings_binary)
             elif self.embeddings_format == 'fasttext':
-                model = KeyedVectors.load_fasttext_format(self.embeddings,
-                                                          binary=self.embeddings_binary)
+                model = fasttext.load_facebook_vectors(self.embeddings)
+                # model = KeyedVectors.load_fasttext_format(self.embeddings,
```
jeremyasapp: please remove the commented code

siddharth-asapp: Done

PR #146, opened 2019-11-19
Github actions

iitzco-asapp:


tox.ini
```diff
@@ -1,16 +1,3 @@
...
-
-# For reference on how to add more features to tox:
-# https://blog.ionelmc.ro/2015/04/14/tox-tricks-and-patterns/
-
-
-# Because of https://github.com/tox-dev/tox/issues/149
-# requirements will not be installed in the [deps] section
-
-[tox]
-envlist = py36, py37
```
iitzco-asapp: don't we need to leave one? If not idk if its a valid tox file.

PR #151, opened 2019-11-19
remove pdb

jeremyasapp:
This PR removes ipdb so that the user can use whichever debugging method they want to use.


iitzco-asapp:
Lets discuss adding later a new parameter for the automatic debugger, as it could still be useful for core objects where adding a breakpoint is not so accessible.

PR #157, opened 2019-11-25
Multiple additions

iitzco-asapp:
- `Experiment` tracks the user that triggered the experiment.
- The `Experiment` saves the YAML version of itself in the output folder.

### DIFF
--- a/flambe/experiment/experiment.py
+++ b/flambe/experiment/experiment.py
@@ -89,6 +89,10 @@ class Experiment(ClusterRunnable):
     merge_plot: bool
         Display all tensorboard logs in the same plot (per block type).
         Defaults to True.
+    user_provider: Callable[[], str]
+        The logic for specifying the user triggering this
+        Runnable. If not passed, by default it will pick the computer's
+        user.
 
     """
 
--- a/flambe/experiment/experiment.py
+++ b/flambe/experiment/experiment.py
@@ -3,10 +3,9 @@ import os
 import re
 import logging
 from copy import deepcopy
-from typing import Dict, Optional, Union, Sequence, cast
+from typing import Dict, Optional, Union, Sequence, cast, Callable
 from collections import OrderedDict
 import shutil
-import getpass
 import tempfile
 
 from tqdm import tqdm
@@ -107,8 +106,9 @@ class Experiment(ClusterRunnable):
                  env: RemoteEnvironment = None,
                  max_failures: int = 1,
                  stop_on_failure: bool = True,
-                 merge_plot: bool = True) -> None:
-        super().__init__(env)
+                 merge_plot: bool = True,
+                 user_provider: Callable[[], str] = None) -> None:
+        super().__init__(env=env, user_provider=user_provider)
         self.name = name
 
         self.original_save_path = save_path
@@ -644,7 +644,7 @@ class Experiment(ClusterRunnable):
             The user as a string.
 
         """
-        return self.env.local_user if self.env else getpass.getuser()
+        return self.env.local_user if self.env else self.user_provider()
 
     def _dump_experiment_file(self) -> None:
         """Dump the experiment YAML representation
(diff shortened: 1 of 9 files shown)


tests/unit/experiment/test_experiment.py
```diff
@@ -0,0 +1,47 @@
...
+    return wrapped
+
+
+@mock.patch('flambe.experiment.experiment.getpass.getuser')
+def test_get_user(mock_user, get_experiment):
+    mock_user.return_value = 'foobar'
+
+    exp = get_experiment()
+    assert exp.get_user() == 'foobar'
+    mock_user.assert_called_once()
```
ciguaran-asapp: This assertion adds no value

jeremyasapp:
Can you explain how this PR works? What happens when running locally? Also aren't you assuming that all cluster implementations ave a user? Is that a valid assumption for the SSHCluster?

flambe/experiment/experiment.py
```diff
@@ -629,3 +634,27 @@ def parse(self) -> None:
...
+    def get_user(self) -> str:
+        """Get the user that triggered this experiment.
+
+        Returns
+        -------
+        str:
+            The user as a string.
+
+        """
+        return self.env.local_user if self.env else getpass.getuser()
```
ciguaran-asapp: Inject get.pass getuser() as a user_provider and you don't need to use patching in the test. You can leave it as default parameter so that there's no need to specify it in the yaml, but you also leave the posibility of external users setting their own users

tests/unit/experiment/test_experiment.py
```diff
@@ -0,0 +1,47 @@
+from flambe.experiment import Experiment
+from flambe.runnable import RemoteEnvironment
+
+import mock
+import pytest
+
+
+@pytest.fixture
+def get_experiment():
+    def wrapped(**kwargs):
```
ciguaran-asapp: Why this kwargs parsing is necessary at test level? Just create an Experiment directly

iitzco-asapp:
> Can you explain how this PR works? What happens when running locally? Also aren't you assuming that all cluster implementations ave a user? Is that a valid assumption for the SSHCluster?

This is the user that launched the experiment. When running locally, the process owner will be picked. When running remotely, the original user that triggered the experiment gets tracked in the environment.

tests/unit/experiment/test_experiment.py
iitzco-asapp: I expect the `test_experiment` file to grow soon. Adding a fixture at this point is good practice so that future tests can use it!

ciguaran-asapp: Yes but why the kwargs?

iitzco-asapp: This is for the user of the fixture to add the attributes that he wants (all other will receive dummy values). The `**kwargs` avoids copying all Experiment parameters.

flambe/experiment/experiment.py
iitzco-asapp: Agreed. Done.

flambe/runnable/runnable.py
```diff
@@ -24,11 +25,13 @@ class Runnable(MappedRegistrable):
         'config["AWS"]["ACCESS_KEY"]'
 
     """
-    def __init__(self, **kwargs) -> None:
+    def __init__(self, user_provider: Callable[[], str] = None, **kwargs) -> None:
```
ciguaran-asapp: Why None? can't the default be DEFAULT_USER_PROVIDER?

tests/unit/experiment/test_experiment.py
```diff
@@ -0,0 +1,48 @@
...
+            user=kwargs.get('user', 'ubuntu'),
+            local_user=kwargs.get('local_user', 'some_user'),
+        )
+        return env
+    return wrapped
+
+
+def test_get_user(get_experiment, get_env):
+    exp = get_experiment(user_provider=lambda: "foobar")
+    assert exp.get_user() == 'foobar'
```
ciguaran-asapp: This is an easy to understand, zero boilerplate test!

flambe/runnable/runnable.py
iitzco-asapp: Fixed.

PR #166, opened 2019-12-01
v0.5 Model, Trainer refactor

jeremyasapp:
There are three core changes in this PR:

1) A refactor of the Trainer object, now called a `Training` task. The idea is to provide users with a single entry point to the software being to implement the missing methods of the `Training` abstract interface. Users are welcome to use the fields, samplers, etc.. that flambe provides, but this interface ultimately provides more room for customization by having everything in one object.

- build_model: constructs the model
- build_optimizers: dictionary pointing from string to optimizer
- build_schedulers: dictionary pointing from string to scheduler
- sample_batches: yields batches of data
- train_step: takes a batch and returns the loss
- val_step: takes all validation batches and returns a set of evaluation metrics
- val_metric: selects a single metric for validation

One of the thing I'm unhappy about in this design is the build_x methods. They are the way they are because I was trying to hide some boilerplate code from the users (namely fp16 and DataParallel) which modify the model but cannot be executed in place. This results in potentially a lot of private attributes used to construct the model, and is generally kind of ugly. Would appreciate suggestions. One alternate design would be to use a callback object of some sort, which Training would leverage.

2) The Encoder interface. Simply expands on the Module object to provide `input_dim` and `output_dim` methods. I'm considering whether we should make these optional methods on flambe.nn.Module instead of adding a new class. These methods provide a consistent way of accessing the size of the dimension that other objects may care about.

3) Optimizers and Scheduler can be instantiated without parameters. This makes the config cleaner, and is generally something I disliked in the original PyTorch interface. Again welcoming suggestions here, as it required writing down every Optimizer in PyTorch..


flambe/nn/module.py
```diff
@@ -120,3 +122,44 @@ def clip_gradient_norm(self, threshold: float):
         # Only compute over parameters that are being trained
         parameters = filter(lambda p: p.requires_grad and p.grad is not None, self.parameters())
         nn.utils.clip_grad_norm_(parameters, threshold)
+
+
+class Encoder(Module):
```
nmatthews-asapp: I think this name is a little confusing because of our history using this word for different things.

flambe/task/training.py
```diff
@@ -0,0 +1,447 @@
...
+    that the internal repesentation of input examples, and of a batch
+    can be completely generics:
+
+    - ``build_model``: constuct the nn.Module to train. The model
+        is passed to the other methods to execute computation.
+    - ``build_optimizers``: given the model, create the optimizers
+        to use to optimize the model. Should be returned as a dict.
+    - ``build_schedulers``: given the optimizers, create the learning
+        rate schedulers. Should be returns as a dict.
+    - ``sampler``: takes an iterable of example data points and returns
```
peterAsapp: it seems like this is called `sample_batches` in the code.

flambe/dataset/tabular.py
```diff
@@ -217,7 +217,9 @@ def __init__(self,
         self.val_view: Optional[DataView] = None
         self.test_view: Optional[DataView] = None
 
-    def _set_transforms(self, transform: Dict[str, Union[Field, Dict]]) -> None:
+    def _set_transforms(self,
+                        transform: Dict[str, Union[Field, Dict]],
+                        do_setup: bool = True) -> None:
```
iitzco-asapp: Add documentation worth it?

flambe/metric/metric.py
```diff
@@ -12,24 +12,11 @@ class Metric(Component):
...
-
-        Returns
-        -------
-        torch.Tensor
-            The computed metric
-
-        """
+    @typing.no_type_check
+    @abstractmethod
+    def compute(self, *args, **kwargs):
```
iitzco-asapp: This is a strong change. Can't we relax the typing but not remove them completely? For example, are we always returning `Tensor` right?

flambe/nlp/classification/__init__.py
```diff
@@ -1,5 +1,6 @@
 from flambe.nlp.classification.datasets import SSTDataset, TRECDataset, NewsGroupDataset
-from flambe.nlp.classification.model import TextClassifier
+from flambe.nlp.classification.task import TextClassification, TextClassifier
```
iitzco-asapp: This is a little bit confusing... is `TextClassifier` a task?

flambe/nlp/classification/task.py
```diff
@@ -0,0 +1,204 @@
...
+from flambe.sampler import BaseSampler
+from flambe.nn import Module, Encoder, Embedder, Embeddings
+from flambe.metric import Accuracy
+from flambe.field import TextField, LabelField
+from flambe.task import Training
+from flambe.optim.optimizer import Optimizer
+from flambe.optim.lr_scheduler import LRScheduler
+
+
+class TextClassifier(Encoder):
```
iitzco-asapp: As I said previously, I would put `TextClassifier` in a `model` module.

flambe/nlp/classification/task.py
```diff
@@ -0,0 +1,204 @@
...
+        eval_scheduler : Optional[LRScheduler], optional
+            A learning rate scheduler to call on every validation step.
+        text_field : Optional[TextField], optional
+            A custom text field to apply to the text inputs.
+        label_field : Optional[LabelField], optional
+            A custom label field to apply to the label inputs.
+        embedder : Optional[Embedder], optional
+            A custom embedder. Overrides ``encoder`` and ``pooling``.
+        model : Optional[TextClassifier], optional
+            A custom model. Overrides ``encoder`` and ``embedder``.
```
iitzco-asapp: I would provide factories to avoid confusion related to this. For example:

```
@staticmethod
def from_model():
   ...
```

flambe/nlp/classification/task.py
```diff
@@ -0,0 +1,204 @@
...
+
+        self.dataset = dataset
+        self.text_field = text_field
+        self.label_field = label_field
+        self.dataset._set_transforms({'text': text_field, 'label': label_field}, do_setup=False)
+
+        if encoder is None and embedder is None and model is None:
+            raise ValueError("At least one of encoder, embedder or model must be provided.")
+        if model is None and embedder is None and encoder is not None and pooling is None:
+            raise ValueError("Must provide a pooling stratgey.")
```
iitzco-asapp: The factories would avoid a lot of these checks

flambe/nlp/language_modeling/task.py
```diff
@@ -0,0 +1,259 @@
...
+            The optimizer to use. Should be provided for training.
+        iter_scheduler : Optional[LRScheduler], optional
+            A learning rate scheduler to call on every training step.
+        eval_scheduler : Optional[LRScheduler], optional
+            A learning rate scheduler to call on every validation step.
+        text_field : Optional[TextField], optional
+            A custom text field to apply to the text inputs.
+        embedder : Optional[Embedder], optional
+            A custom embedder. Overrides ``encoder`` and ``pooling``.
+        model : Optional[Encoder], optional
```
iitzco-asapp: Same here with the factories.

flambe/nlp/language_modeling/task.py
```diff
@@ -0,0 +1,259 @@
...
+        self.tie_weights = tie_weights
+        self.loss_fn = nn.CrossEntropyLoss(reduction='none')
+        self.metric_fn = BPC() if use_bpc else Perplexity()
+        self.metric_name = 'bpc' if use_bpc else 'perplexity'
+
+        self.optimizer = optimizer
+        self.eval_scheduler = eval_scheduler
+        self.iter_scheduler = iter_scheduler
+
+    def build_model(self):
```
iitzco-asapp: There is a lot of typing missing... not only in this file!

flambe/nlp/transformers/xl.py
```diff
@@ -32,3 +32,9 @@ class TransfoXLEmbedder(TransformerEmbedder):
     """
 
     _cls = pt.TransfoXLModel
+
+    def tie_weights(self, linear):
```
iitzco-asapp: Typing, as with all `tie_weights` methods

flambe/nn/cnn.py
```diff
@@ -3,24 +3,24 @@
...
+               pooling: Optional[nn.Module] = None,
                batch_norm: Optional[nn.Module] = None) -> nn.Module:
     """Return a convolutional block.
 
     """
 
     mods = [conv_mod]
 
-    if pooling:
+    if pooling is not None:
```
iitzco-asapp: Why go this path?

flambe/nn/cnn.py
```diff
@@ -3,24 +3,24 @@
...
         mods.append(pooling)
 
-    if batch_norm is None:
+    if batch_norm is not None:
```
iitzco-asapp: Good catch.

flambe/nn/cnn.py
```diff
@@ -122,18 +122,44 @@ def __init__(self,
                     raise ValueError("Kernel size tuple should have same length as conv_dim")
 
             layer = conv_block(
-                conv(prev_c, c, k, stride, padding),
+                conv(prev_c, c, k, stride, padding),  # type: ignore
                 activation,
-                pooling,
                 dropout,
+                pooling,
```
iitzco-asapp: Pooling after dropout will alter the dropout effect... at least in my intuition. Is there a reason to do this?

flambe/nn/module.py
```diff
@@ -120,3 +122,44 @@ def clip_gradient_norm(self, threshold: float):
...
+    def input_dim(self) -> int:
+        """Get the size of the last dimension of an input.
+
+        Returns
+        -------
+        int
+            The size of the last dimension of an input.
+
+        """
+        pass
```
iitzco-asapp: do we want to `pass` here or `raise NotImplementedError`? Same below

flambe/nn/pooling.py
```diff
@@ -32,7 +32,7 @@ def forward(self,
 class LastPooling(Module):
     """Get the last hidden state of a sequence."""
 
-    def forward(self,
+    def forward(self,  # type: ignore
```
iitzco-asapp: Typing was working before... why removing it now?

flambe/nn/rnn.py
```diff
@@ -83,9 +80,9 @@ def __init__(self,
         self.enforce_sorted = enforce_sorted
         if rnn_type in ['lstm', 'gru']:
             if kwargs:
-                logger.warn(f"The following '{kwargs}' will be ignored " +
-                            "as they are only considered when using 'sru' as " +
-                            "'rnn_type'")
+                warnings.warn(f"The following '{kwargs}' will be ignored " +
```
iitzco-asapp: Lets try if this doesn't break our logging mechanism (ie that the warnings appear in our logs)

flambe/optim/lr_scheduler/noam.py
```diff
@@ -1,21 +1,23 @@
...
 
     """
 
     def __init__(self,
-                 optimizer,
-                 warmup: int,
-                 d_model: int):
+                 optimizer: Optional[Optimizer] = None,
+                 warmup: int = 100,
+                 d_model: int = 768) -> None:
```
iitzco-asapp: extract into constants maybe? Why `768`?

flambe/optim/optimizer/optimizer.py
```diff
@@ -0,0 +1,73 @@
...
+class LBFGS(Optimizer, torch.optim.LBFGS):
+    _cls = torch.optim.LBFGS
+
+
+class RMSprop(Optimizer, torch.optim.RMSprop):
+    _cls = torch.optim.RMSprop
+
+
+class Rprop(Optimizer, torch.optim.Rprop):
+    _cls = torch.optim.Rprop
```
iitzco-asapp: Does it make sense to create them dynamically? I imagine this will be hard to maintain.

iitzco-asapp: Maybe we should always use `Optimizer` and pass the `cls` as a parameter. Then:

```
!Optimizer:
    cls: !torch.SGD
    params: ...
```

flambe/task/legacy.py
```diff
@@ -1,4 +1,5 @@
 import math
```
iitzco-asapp: Any reason to keep the existing code? If this is a backwards incompatible change, lets throw the current implementation so that we don't maintain it.

flambe/task/script.py
```diff
@@ -66,5 +66,20 @@ def run(self) -> bool:
...
+        a unique scalar representing the current performance on the
+        task, and to compare against other variants of the task.
+
+        Returns
+        -------
+        Optional[float]
+            The metric to compare different variants of your Component
+
+        """
+        return None
```
iitzco-asapp: What does it mean to return `None`?

flambe/vision/classification/task.py
```diff
@@ -0,0 +1,124 @@
...
+    def output_dim(self) -> int:
+        """Get the size of the last dimension of an output."""
+        return self.output_size
+
+    def forward(self, data):  # type: ignore
+        """Run a forward pass from shape (B x H) to (B x O)."""
+        return self.output_layer(self.drop(self.encoder(data).flatten(1)))
+
+
+class ImageClassification(TextClassification):
```
iitzco-asapp: Why inherit from `TextClassification` ? I think this is a typo.

flambe/vision/classification/task.py
```diff
@@ -0,0 +1,124 @@
...
+            The batch size to use during evaluation. Default ``32``.
+        optimizer : Optional[Optimizer], optional
+            The optimizer to use. Should be provided for training.
+        iter_scheduler : Optional[LRScheduler], optional
+            A learning rate scheduler to call on every training step.
+        eval_scheduler : Optional[LRScheduler], optional
+            A learning rate scheduler to call on every validation step.
+        label_field : Optional[LabelField], optional
+            A custom label field to apply to the label inputs.
+        model : Optional[ImageClassifier], optional
```
iitzco-asapp: Factories!

jeremyasapp:
@iitzco-asapp 

1. Not everything inherits from Encoder
2. I was having issues with mypy, may be resolved, I'll have to check
3. I ultimately felt like inheritance provided a better user experience here.
4. Testing will be done the same way as it is done now.

flambe/metric/loss/distillation.py
```diff
@@ -0,0 +1,78 @@
...
+            Defaults to 0.5.
+        temperature: float
+            Temperature applied to the student and teacher
+            distributions. Defaults to 1.
+        objective_fn: Callable, Optional
+            The loss function to use over the logits for the student.
+            By default, uses Pytorch's ``CrossEntropyLoss``.
+
+        """
+        self.objective_fn = objective_fn or torch.nn.CrossEntropyLoss(reduction='none')
```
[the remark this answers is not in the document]
jeremyasapp: this is a sensible default

flambe/nn/transformer_sru.py
```diff
@@ -423,7 +495,8 @@ def __init__(self,
         """
         super().__init__()
 
-        self.self_attn = nn.MultiheadAttention(d_model, nhead, dropout=dropout)
+        self.d_model = d_model
```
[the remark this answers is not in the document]
jeremyasapp: meant for consistency with PyTorch. We could consider not following their naming

flambe/dataset/tabular.py
jeremyasapp: Not if it remains private. I do think this code needs a bit of refactoring though

flambe/metric/metric.py
jeremyasapp: The better question is: why restrict the type?

flambe/nlp/classification/__init__.py
jeremyasapp: I can put them in separate modules. TextClassification is the task.

flambe/nlp/classification/task.py
jeremyasapp: i ll make that change

flambe/nlp/classification/task.py
jeremyasapp: I think you mean classmethod, but yeah, we could do that. I don't think it's that much better though

flambe/nlp/classification/task.py
jeremyasapp: that's fair

flambe/nlp/language_modeling/task.py
jeremyasapp: Will fix

flambe/nlp/classification/task.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __init__(self,
+                 dataset: TabularDataset,
+                 encoder: Optional[Encoder] = None,
+                 pooling: Optional[Module] = None,
+                 dropout: float = 0,
+                 build_vocabularies: bool = True,
+                 train_batch_size: int = 32,
+                 val_batch_size: int = 32,
+                 embedding_args: Dict[str, Any] = None,
```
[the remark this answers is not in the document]
jeremyasapp: why?

flambe/nlp/fewshot/prototypical.py
```diff
@@ -0,0 +1,182 @@
...
+        super().__init__(dataset, **kwargs)
+        self.distance = distance
+        self.n_episodes = n_episodes
+        self.n_classes = n_classes
+        self.n_support = n_support
+        self.n_query = n_query
+        self.balance_query = balance_query
+        self.detach_mean = detach_mean
+
+    def build_model(self):
```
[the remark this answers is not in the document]
jeremyasapp: yeah i don't like the build_model method anymore

flambe/nn/cnn.py
jeremyasapp: pooling is a Module. I generally prefer using `is not None` on non built-in objects, as they may or may not implement the boolean method

flambe/nn/cnn.py
jeremyasapp: Yes, i don't want the last returned vector to have dropout, I want the dropout to remain somewhat internal. So if someone wants to add another dropout between the result of the encoder and the final projection (say for a classifier), they can

flambe/nn/module.py
jeremyasapp: always pass on abstractmethods

flambe/nn/module.py
jeremyasapp: Suggestions?

flambe/optim/lr_scheduler/scheduler.py
```diff
@@ -0,0 +1,68 @@
...
+
+class LRScheduler(Component, lr_scheduler._LRScheduler):
+
+    _cls: Type[lr_scheduler._LRScheduler]
+
+    def __init__(self, optimizer=None, **kwargs):
+        self.initialized = False
+        if optimizer is not None:
+            self.initialized = True
+            self._cls.__init__(self, optimizer=optimizer, **kwargs)
```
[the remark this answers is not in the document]
jeremyasapp: yes

flambe/optim/lr_scheduler/scheduler.py
```diff
@@ -0,0 +1,68 @@
...
+        else:
+            self.kwargs = kwargs
+
+    def initialize(self, optimizer):
+        if self.initialized:
+            raise ValueError("Scheduler already initialized with optimizer.")
+
+        self.initialized = True
```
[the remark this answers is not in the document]
jeremyasapp: true

flambe/optim/optimizer/optimizer.py
jeremyasapp: We have no way of passing just the class at the moment in YAML, so I didn't consider this as an option. I agree it would be nice if there was a better way of loading these guys dynamically. I'm just not sure how

flambe/task/legacy.py
jeremyasapp: Like we discussed offline, we keep the old code for allowing people to smoothly transition

flambe/task/script.py
jeremyasapp: I was thinking we could allow returning None which means the Task cannot be used with `reduce` or certain hyperparameter search aglrotihms

flambe/vision/classification/task.py
jeremyasapp: it saves some code, but I could build a Classification task instead which they both inherit from

flambe/task/training.py
```diff
@@ -0,0 +1,447 @@
...
+        -------
+        torch.Tensor
+            Should contain at least one element being the loss for the
+            batch. The recommended default key is ``loss``.
+
+        """
+        pass
+
+    @abstractmethod
+    def val_step(self, model: nn.Module, batches: Iterable[Batch]) -> Dict[str, float]:
```
ciguaran-asapp: If these methods would be objects that are received by the training, then that would help code reuse. For example, if several tasks share a way of doing validation steps, each of them needs to inherit and override, but it will be tricky to share that code among all those clients.

flambe/task/training.py
```diff
@@ -0,0 +1,447 @@
...
+    - ``eval_step``: takes an iterable over batches and returns the
+        metric to use for validation. Higher should indicate better.
+
+    """
+
+    def __init__(self,
+                 max_epoch: float = 1.0,
+                 max_iter: Optional[int] = None,
+                 gradient_accumulation: int = 1,
+                 max_grad_norm: Optional[float] = None,
```
ciguaran-asapp: These two parameters solve the problem of grandiet clipping, can't they be extracted out and have a GrandietClipper()? Different implementations of that clipper may do it by gradient norm, or abs val, or both

flambe/task/training.py
```diff
@@ -0,0 +1,447 @@
...
+                 max_grad_abs_val: Optional[float] = None,
+                 validation_freq: float = 0.01,
+                 eval_only: bool = False,
+                 eval_on_start: List[str] = ['val'],
```
ciguaran-asapp: These three variables are unused in this file?

flambe/task/training.py
jeremyasapp: Every single one being objects would be overkill and require inspecting many objects to get going, which is against the spirit of this change (you're essentially describing the status quo :p ). but one alternative is to have Training receive a Task object with the methods

flambe/task/training.py
jeremyasapp: I also think that's a bit overkill. Maybe we can revisit when a user has a specific requirement that justifies that change.

flambe/task/training.py
jeremyasapp: i believe they are used

flambe/task/training.py
ciguaran-asapp: Yes, but the problem is going to be when you try to test this. In order to test an override, since probably the implementations will depend on the class internal state, you will need to exercise a lot of code from the base class. Probably the thing here is that this class handles to many things

jeremyasapp: yeah that's a fair point. But I think there is a good middle ground we can find :)

jeremyasapp:
Closing as this as being reviewed.

PR #168, opened 2019-12-02
Best model cpu jeremyasapp

jeremyasapp:


flambe/learn/train.py
```diff
@@ -267,7 +266,7 @@ def _eval_step(self) -> None:
         sign = (-1)**(self.lower_is_better)
         if self._best_metric is None or (sign * val_metric > sign * self._best_metric):
             self._best_metric = val_metric
-            self._best_model = deepcopy(self.model.state_dict())
+            self._best_model = {k: t.cpu().detach() for k, t in self.model.state_dict().items()}
```
nmatthews-asapp: If you change this code to use the existing dictionary object instead of creating a new one (i.e. modify in place) then I think the bug will go away and your tests will pass.

PR #169, opened 2019-12-02
Added command line `--debug` argument

cle-ros:
Added a command line `-d` / `--debug` argument that overrides the yaml setting.

### DIFF
--- a/flambe/runner/run.py
+++ b/flambe/runner/run.py
@@ -38,6 +38,9 @@ def main(args: argparse.Namespace) -> None:
         print(cl.RA(ASCII_LOGO))
         print(cl.BL(f"VERSION: {flambe.__version__}\n"))
 
+    if args.debug and args.cluster is not None:
+        raise ValueError('Will not run on cluster in debug mode. Please disable debug mode or run locally.')
+
     # Pass original module for ray / pickle
     make_component(torch.nn.Module, only_module='torch.nn')
     # torch.optim.Optimizer exists, ignore mypy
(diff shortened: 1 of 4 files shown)


flambe/runner/run.py
```diff
@@ -93,7 +97,7 @@ def main(args: argparse.Namespace) -> None:
                 else:
                     raise ValueError("Only ClusterRunnables can be executed in a cluster.")
         else:
-            runnable, _ = ex.preprocess(secrets=args.secrets, install_ext=args.install_extensions)
+            runnable, _ = ex.preprocess(secrets=args.secrets, install_ext=args.install_extensions, debug=args.debug)
```
iitzco-asapp: I would prefer the `debug` parameter to be passed to the `run` method of the runnables instead of injecting it through the `preprocess`.

This would involve changing the API of runnable to receive this extra parameter (defaults to `False`) in its `run` method.

iitzco-asapp:
I like this flag better that the one that is used inside the config.

I suggest removing the one in the config completely (that is, remove the `debug` parameter in `Experiment`.

peterAsapp:
I think we should do the same for `resume` also. If you guys agree, I can make an issue.

jeremyasapp:
@peterAsapp I'm not sure i agree about resume since resume is very much an Experiment specific thing. You could argue the same about the `save_path` too

peterAsapp:
mm i see what you mean. my reasoning was that we usually always run an `Experiment` using flambe, but I guess you could run something else by design.

iitzco-asapp:
This feature will be merged in #173

PR #173, opened 2019-12-06
Follow up on #169

iitzco-asapp:
See #169

### DIFF
--- a/flambe/runner/run.py
+++ b/flambe/runner/run.py
@@ -39,7 +39,8 @@ def main(args: argparse.Namespace) -> None:
         print(cl.BL(f"VERSION: {flambe.__version__}\n"))
 
     if args.debug and args.cluster is not None:
-        raise ValueError('Will not run on cluster in debug mode. Please disable debug mode or run locally.')
+        raise ValueError('Will not run on cluster in debug mode. ' +
+                         'Please disable debug mode or run locally.')
 
     # Pass original module for ray / pickle
     make_component(torch.nn.Module, only_module='torch.nn')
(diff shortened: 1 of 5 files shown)


flambe/runner/run.py
```diff
@@ -114,6 +119,8 @@ def main(args: argparse.Namespace) -> None:
                              'when using this flag as it could have undesired effects.')
     parser.add_argument('-s', '--secrets',
                         type=str, default=os.path.join(FLAMBE_GLOBAL_FOLDER, "secrets.ini"))
+    parser.add_argument('-d', '--debug', action='store_true',
+                        help='Enable debug mode, overriding config.yaml setting')
```
nmatthews-asapp: This no longer overrides `config.yaml` setting, right? Also I think we should explain a bit more what debug does, e.g. explicitly say "For an experiment job, Ray will run in a single thread if debug mode is enabled."

iitzco-asapp: Done.

flambe/runner/run.py
```diff
@@ -38,6 +38,10 @@ def main(args: argparse.Namespace) -> None:
         print(cl.RA(ASCII_LOGO))
         print(cl.BL(f"VERSION: {flambe.__version__}\n"))
 
+    if args.debug and args.cluster is not None:
```
jeremyasapp: Are we sure that we want to deactivate all of the `debug` functionalities in remote execution?

iitzco-asapp: It's not clear to me how `debug` can work in a remote execution. Until we figure that out, I suggest `debug` just works locally for any runnable.

jeremyasapp: sounds good

PR #174, opened 2019-12-06
Add warning when content to upload exceeds limit

iitzco-asapp:
(cc @ciguaran-asapp)

When uploading heavy content to the cluster (either an extension or a resource), `flambe` was lacking a warning so that users understand where this delay is coming from.

This PR fixes this.

- [x] Add unit testing for the `get_size_MB` method

### DIFF
--- a/flambe/cluster/cluster.py
+++ b/flambe/cluster/cluster.py
@@ -578,7 +578,11 @@ class Cluster(Runnable):
 
                 size = get_size_MB(c)
                 if size > UPLOAD_WARN_LIMIT_MB:
-                    logger.info(cl.YE(f"Uploading '{c}' ({int(size)} MB) which may take a while; if this is a surprise, double check you want to be transferring this file. Note we automatically sync resources, extensions (and any files in your flambe installation if installed with '-e')"))
+                    logger.info(cl.YE(
+                        f"Uploading '{c}' ({int(size)} MB) which may take a while. " +
+                        "Double check you want to be transferring this file " +
+                        "(note we automatically sync extensions and experiment resources " +
+                        "and potentially the flambe repo if installed in dev mode)"))
 
                 if os.path.isdir(c):
                     if not c.endswith(os.sep):
--- a/flambe/cluster/cluster.py
+++ b/flambe/cluster/cluster.py
@@ -8,6 +8,7 @@ from flambe.compile import yaml
 from flambe.runnable import Runnable
 from flambe.runnable.error import MissingSecretsError
 from flambe.runnable.utils import is_dev_mode
+from flambe.runner.utils import get_size_MB
 from flambe.cluster import const
 from flambe.cluster.instance import errors
 from flambe.cluster import errors as man_errors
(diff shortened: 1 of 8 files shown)


flambe/runner/utils.py
```diff
@@ -26,11 +26,14 @@ def get_folder_size_MB(path: str) -> float:
...
-        for f in filenames:
-            fp = os.path.join(dirpath, f)
-            if os.path.exists(fp) and not os.path.islink(fp):
-                accum += os.path.getsize(fp)
+    if os.path.isdir(path):
+        for dirpath, dirnames, filenames in os.walk(path):
+            for f in filenames:
+                fp = os.path.join(dirpath, f)
+                if os.path.exists(fp) and not os.path.islink(fp):
+                    accum += os.path.getsize(fp)
```
ciguaran-asapp: Do we really need to do a summation here? It shouldn't be necessary to traverse the path.

flambe/runner/utils.py
```diff
@@ -50,6 +53,6 @@ def check_system_reqs() -> None:
 
     # Check if extensions folder is getting big
     extensions_folder = os.path.join(FLAMBE_GLOBAL_FOLDER, "extensions")
-    if get_folder_size_MB(extensions_folder) > WARN_LIMIT_MB:
+    if os.path.exists(extensions_folder) and get_size_MB(extensions_folder) > WARN_LIMIT_MB:
```
ciguaran-asapp: I don't follow why we are logging this in two different places.

flambe/runner/utils.py
iitzco-asapp: There is no standard way of getting the size of a directory (something similar to `du`). `os.walk` seems the way to go here, even the `os.walk` docs has an example on how to do this (https://docs.python.org/3/library/os.html#os.walk)

flambe/runner/utils.py
iitzco-asapp: Good point.

flambe/cluster/cluster.py
```diff
@@ -571,6 +575,11 @@ def send_local_content(self,
             c = os.path.expanduser(c)
             base: str = ""
             if os.path.exists(c):
+
+                size = get_size_MB(c)
+                if size > UPLOAD_WARN_LIMIT_MB:
+                    logger.info(cl.YE(f"Uploading may take time for ({int(size)} MB) '{c}'"))
```
nmatthews-asapp:
```suggestion
                    logger.info(cl.YE(f"Uploading '{c}' ({int(size)} MB) which may take a while; if this is a surprise, double check you want to be transferring this file. Note we automatically sync resources, extensions (and any files in your flambe installation if installed with '-e')"))
```

flambe/cluster/cluster.py
```diff
@@ -571,6 +575,15 @@ def send_local_content(self,
             c = os.path.expanduser(c)
             base: str = ""
             if os.path.exists(c):
+
+                size = get_size_MB(c)
+                if size > UPLOAD_WARN_LIMIT_MB:
+                    logger.info(cl.YE(
+                        f"Uploading '{c}' ({int(size)} MB) which may take a while. " +
+                        "Double check you want to be transferring this file " +
+                        "(note we automatically sync extensions and experiment resources " +
```
iitzco-asapp:
```suggestion
                        "(note we automatically sync extensions, experiment resources " +
```

flambe/cluster/cluster.py
```diff
@@ -571,6 +575,15 @@ def send_local_content(self,
...
+                        "and potentially the flambe repo if installed in dev mode)"))
```
iitzco-asapp:
```suggestion
                        "and the flambe repo if installed in dev mode)"))
```

PR #175, opened 2019-12-10
Adding tests for python 3.7

iitzco-asapp:
- `tox` now must execute test suite in `3.7`
- CCI was fixed so that it can install requirements in `3.7`

**The python:3.6 image we are using already contains both python 3.6 and python 3.7**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,9 @@
+python3: &python3
+    command: |
+        apt-get update
+        apt-get install -y python3-dev
+
+
 version: 2
 jobs:
   unit_tests:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -28,6 +32,10 @@ jobs:
         - image: python:3.6
```
jeremyasapp: isn't this a problem?

iitzco-asapp: This image also provides `3.7`

iitzco-asapp:
> Just wanted to clarify - this will have python 3.7 used to run our tests only? I realize the image has both installed, but how are we testing / what are we officially supporting?

`tox` FTW!

iitzco-asapp:
> There is a lot of overlap in the config.yml besides what you put in the anchor. Was there a reason why you limited it to just those 2 python 3 install steps?

`python3` anchor is just to install python3.7. It is not needed when publishing docs or the package. Naming is not the best I think... I will update.

PR #180, opened 2019-12-13
GCP Integration

yukw777:
A draft PR for GCP integration into flambe. `libcloud` was extremely easy to use, and I think it'd allow us to rapidly add more cloud services to Flambe. We should probably think about moving `AWSCluster` over to `libcloud` also.

`libcloud` has this concept of `Deployment`, a series of tasks you can pass in when launching an instance. I think we can replace a lot of the code we use to set up an instance by utilizing this. We'd still use paramiko to run commands after the instances are ready though. I couldn't do this for this PR as their GCP driver for some reason doesn't support this, and I've sent a PR to fix this (https://github.com/apache/libcloud/pull/1387). I'll update the PR when a new release with my changes comes out.

Here is the TODO list:
- [ ] documentation
- [ ] unit tests
- [x] ~automatic shutdown of instances~ Unfortunately, GCP doesn't let you set up actions based on metrics alerts, so we'd need another solution for this. Out-of-scope for this PR.
- [ ] move custom commands to a script file.
- [x] tags for instances
- [ ] disk size (default 100 gb) -> waiting on this: https://github.com/apache/libcloud/pull/1388
- [x] disk type (standard vs ssd)

Let me know if there is anything else I need to do to get this merged.


flambe/cluster/const.py
```diff
@@ -3,7 +3,7 @@
 
 # SSH connection settings
 RETRY_DELAY = 1
-RETRIES = 60
+RETRIES = 300
```
yukw777: It takes longer for GCP GPU instances using their standard pytorch image to become available as the image for some reason doesn't come with the Nvidia drivers and instances need to install them when they come up. So I had to increase the retry number. It typically takes about 2-3 minutes. The workaround is either to create an image with the nvidia drivers pre-installed, or use another publicly available image that comes with the driver.

iitzco-asapp:
THis is awesome. @yukw777 tells me when its ready for final review.

jeremyasapp:
We will probably want to close this since ray provides a GCP integration already. We can also consider libcloud though, in which case some of this code may be recycled but unclear how much at the moment.

jeremyasapp:
Closing as a cluster refactor PR is on the way

PR #184, opened 2019-12-20
get batch inside try-except

cdfox-asapp:


flambe/learn/train.py
```diff
@@ -184,45 +184,51 @@ def _train_step(self) -> None:
...
+        try:
+            with torch.enable_grad():
+                for i in range(self.iter_per_step):
+                    print(f"iter {i}")
+                    # Zero the gradients and clear the accumulated loss
+                    self.optimizer.zero_grad()
+                    accumulated_loss = 0.0
+                    for _ in range(self.batches_per_iter):
+                        # Get next batch
+                        batch = next(self._train_iterator)
```
jeremyasapp: I think the correct solution to this problem is to add the try except here, and reload the iterator:

```
try:
    batch = next(self._train_iterator)
except StopIteration:
    self._train_iterator = self.train_sampler.sample()
```

cdfox-asapp: ok, I pulled the self.train_sampler.sample call into a separate method, since it needs the data and n_epochs as well, so that I could call that method from both here and the __init__

cdfox-asapp: something broke -- I'll be back week of Christmas

jeremyasapp: Yeah you're just missing the batch! 

```
self._create_train_iterator()
batch = next(self._train_iterator)
```

jeremyasapp: After line 202

cdfox-asapp:
going to close this and re-open with a better branch name

PR #196, opened 2020-01-07
field accept vocabs

jeremyasapp:
`TextField` and `LabelField` now accept vocabularies as input, either as a list or a file path.

- Still requires some testing
- Could be expanded by adding support for freezing the vocab when one is provided, or allow it to expand


flambe/field/label.py
```diff
@@ -27,21 +27,29 @@ def __init__(self,
...
+            encountered during preprocessing. The list can also be
+            provided as a file with a label on each line.
 
         """
         self.one_hot = one_hot
         self.multilabel_sep = multilabel_sep
         self.tokenizer = LabelTokenizer(multilabel_sep=self.multilabel_sep)
 
         if labels is not None:
+            if isinstance(labels, str):
```
ciguaran-asapp: I'd use a registered factory to avoid coupling the implementation to a specific format/way of building Also, it will allow easier testing, since you could test separately the way its built from the internal operations

jeremyasapp: I agree in spirit, but do you think it's very valuable here? I feel like the factory method will be so trivial it might not be worth it

jeremyasapp: I hesitated a bit before going for this implementation, but I'm ok with going the classmethod route

ciguaran-asapp: Well, apart of helping the developers in the future (they could add new ways of building the thing without needing to change the code), it could help users right?  The code would be more explicit on the fact that a path for a file is being used

jeremyasapp: yeah i ll change it!

cle-ros:
How does this relate to https://github.com/asappresearch/flambe/pull/257 ?
@iitzco-asapp @jeremyasapp

PR #198, opened 2020-01-09
Extra validation metrics are provided as key-value mapping rather tha…

kball-asapp:
…n a List

This allows you to have multiple metrics of the same class with different configurations. When provided as a list, they're written to tensorboard using the class name, so metrics of the same class are overwritten. A mapping allows each metric object to have a unique name.

### DIFF
--- a/flambe/learn/train.py
+++ b/flambe/learn/train.py
@@ -93,7 +93,7 @@ class Trainer(Component):
         max_grad_abs_val: float, optional
             Maximum absolute value of all gradient vector components
             after clipping.
-        extra_validation_metrics: Optional[List[Metric]]
+        extra_validation_metrics: Optional[Dict[str, Metric]]
             A list with extra metrics to show in each step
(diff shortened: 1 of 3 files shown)


flambe/learn/train.py
```diff
@@ -296,8 +296,8 @@ def _eval_step(self) -> None:
         log(f'{tb_prefix}Validation/Loss', val_loss, self._step)
         log(f'{tb_prefix}Validation/{self.metric_fn}', val_metric, self._step)
         log(f'{tb_prefix}Best/{self.metric_fn}', self._best_metric, self._step)  # type: ignore
-        for metric in self.extra_validation_metrics:
-            log(f'{tb_prefix}Validation/{metric}',
+        for metric_name, metric in self.extra_validation_metrics.items():
+            log(f'{tb_prefix}Validation/{metric_name}',
```
iitzco-asapp: 👍

PR #199, opened 2020-01-10
Goodbye automatic install feature

iitzco-asapp:
This PR removes completely the automatic install feature from flambe. **We decided this feature tries to solve a problem that is not inside flambe's scope, and the current implementation provides more problems than solutions**.

Moving forward, the user will be responsible for installing extensions before using the runnables. Additionally, we are evaluating building supporting tools to deal with the extensions' installation process easier, but they won't be part of the flambe package.


jeremyasapp:
This is gone already in the 0.5 refactor. I would just kill this PR

PR #201, opened 2020-01-14
Fix '-f' bug in first time remote experiments

iitzco-asapp:
There is a bug currently that doesn't allow to run remote experiments if the `-f` flag is not provided (even if its the first time the cluster is being used).

This PR solves this bug.

### DIFF
--- a/flambe/experiment/experiment.py
+++ b/flambe/experiment/experiment.py
@@ -26,6 +26,7 @@ from flambe.experiment.options import ClusterResource
 from flambe.runnable import RemoteEnvironment
 from flambe.runnable import error
 from flambe.runnable import utils as run_utils
+from flambe.runner.utils import get_files
 from flambe.experiment.progress import ProgressState
 from flambe.experiment.tune_adapter import TuneAdapter
 from flambe.logging import coloredlogs as cl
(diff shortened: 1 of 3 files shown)


flambe/experiment/experiment.py
```diff
@@ -194,11 +195,8 @@ def run(self, force: bool = False, verbose: bool = False, debug: bool = False, *
 
         # Check if save_path/name already exists + is not empty
         # + force and resume are False
-        if (
-            os.path.exists(self.full_save_path) and
-            os.listdir(self.full_save_path) and
-            not self.resume and not force
-        ):
+        if not self.resume and not force and os.path.exists(self.full_save_path) \
+                and list(get_files(self.full_save_path)):
```
jeremyasapp: use `os.listdir` instead

iitzco-asapp: `os.listdir` is non recursive and in remote clusters there is an empty folder created in the output folder (called `_resources`), so `os.listdir` resolves to `True` in this conditional... causing the bug I describe above.

This fix uses a recursive implementation where it checks if any file exists. So if a remote execution is taking place, then even if there is a empty folder, this will resolve to `False`.

jeremyasapp: how about `os.walk`?

peterAsapp: or `glob` maybe?

iitzco-asapp: `get_files` uses `os.walk` internally @jeremyasapp

PR #203, opened 2020-01-15
Loss func train eval call

cle-ros:
Slight change to the trainer so that both the eval loop and the train loop use the _compute_batch function (successor of _compute_loss). This way, subclassing for exotic loss functions becomes easier because one does not also need to overwrite the eval steps


iitzco-asapp:
I like this. @cle-ros can you update the branch so that we can merge this?

cle-ros:
> I like this. @cle-ros can you update the branch so that we can merge this?

Merged the latest master.

flambe/optim/noam.py
```diff
@@ -53,4 +53,4 @@ def lr_lambda(self, step: int) -> float:
             if step > self.warmup:
                 return 1. / (self.d_model ** 0.5) / (step ** 0.5)
             else:
-                return step / (self.d_model ** 0.5) / (self.warmup ** 1.5)
+                return (step + 1) / (self.d_model ** 0.5) / (self.warmup ** 1.5)
```
jeremyasapp: i took the formula from the original implementation. Not sure there should be a one here

flambe/learn/train.py
```diff
@@ -245,20 +261,20 @@ def _aggregate_preds(self, data_iterator: Iterator) -> Tuple[torch.Tensor, torch
...
+        preds, targets, loss = [], [], []
         for batch in data_iterator:
-            batch = self._batch_to_device(batch)
-            pred, target = self.model(*batch)
+            pred, target, batch_loss = self._compute_batch(batch)
+            loss.append(batch_loss.item())
             preds.append(pred.cpu())
             targets.append(target.cpu())
-
+        loss = sum(loss)/len(loss)
```
jeremyasapp: loss.mean()?

cle-ros: the losses there are a list of floats. I can change it to be a list of detached loss tensors, but then I'd need to do a concatenation first.

flambe/optim/noam.py
cle-ros: Yes. But that sets the learning rate to zero when the step equals zero (which I think I've seen happen). Given that we also have `step > warmup` instead of `step >= warmup`, I think that this may be more consistent. I'm happy to change it back, though - it does not really matter that much.

flambe/learn/train.py
jeremyasapp: Need a space between the operator /. Probably why the test is failing

jeremyasapp:
Was merged in #226

PR #212, opened 2020-01-17
Hotfix save torch state 

nmatthews-asapp:
Currently there is a bug where we miss some torch state in specific architectures. This PR adds tests for those architectures and fixes the bugs.

TODO
- [x] clean up code
- [x] verify solves target issues
- [x] check backwards compatibility with old save files

### DIFF
--- a/flambe/compile/component.py
+++ b/flambe/compile/component.py
@@ -932,7 +932,6 @@ class Component(Registrable):
             Why the exception is raised.
 
         """
-        print(f"Calling hook on class {type(self)}")
         warn_use_state = False
         if FLAMBE_DIRECTORIES_KEY not in state_dict._metadata:
             state_dict._metadata[FLAMBE_DIRECTORIES_KEY] = set()
@@ -961,9 +960,9 @@ class Component(Registrable):
         # modules, parameters or buffers
         # torch.optim.Optimizer does exist so ignore mypy
         for name, attr in self.__dict__.items():
+            current_path = prefix + name
             if isinstance(attr, Component) and not isinstance(attr, (
                     torch.optim.Optimizer, torch.optim.lr_scheduler._LRScheduler)):  # type: ignore
-                current_path = prefix + name
                 # If self is not nn.Module, need to recurse because
                 # that will not happen elsewhere
                 # If self *is* an nn.Module, don't need to recurse on
--- a/flambe/compile/component.py
+++ b/flambe/compile/component.py
@@ -981,8 +981,8 @@ class Component(Registrable):
             # nn.Module, because otherwise this hook is being called
             # via nn.Module.state_dict, and will already recurse to
             # all children modules
-            if ((not isinstance(self, torch.nn.Module)) and isinstance(attr, torch.nn.Module)
-                    and (not isinstance(attr, Component))):
+            if not isinstance(self, torch.nn.Module) and isinstance(attr, torch.nn.Module) \
+                    and not isinstance(attr, Component):
                 state_dict = attr.state_dict(destination=state_dict,
                                              prefix=current_path + STATE_DICT_DELIMETER,
                                              keep_vars=state_dict._metadata[KEEP_VARS_KEY])
--- a/flambe/compile/component.py
+++ b/flambe/compile/component.py
@@ -929,6 +929,7 @@ class Component(Registrable):
             Why the exception is raised.
 
         """
+        print(f"Calling hook on class {type(self)}")
         warn_use_state = False
         if FLAMBE_DIRECTORIES_KEY not in state_dict._metadata:
             state_dict._metadata[FLAMBE_DIRECTORIES_KEY] = set()
@@ -973,8 +974,16 @@ class Component(Registrable):
                                                 prefix=current_path + STATE_DICT_DELIMETER,
                                                 keep_vars=state_dict._metadata[KEEP_VARS_KEY])
                 state_dict._metadata[FLAMBE_DIRECTORIES_KEY].add(current_path)
-        # Iterate over modules to make sure Component
-        # nn.Modules are added to flambe directories
+            # Iterate over modules to make sure NON-Component
+            # nn.Modules' state is added. Only needed if self is not
+            # nn.Module, because otherwise this hook is being called
+            # via nn.Module.state_dict, and will already recurse to
+            # all children modules
+            if ((not isinstance(self, torch.nn.Module)) and isinstance(attr, torch.nn.Module)
+                    and (not isinstance(attr, Component))):
+                state_dict = attr.state_dict(destination=state_dict,
+                                             prefix=current_path + STATE_DICT_DELIMETER,
+                                             keep_vars=state_dict._metadata[KEEP_VARS_KEY])
         state_dict._metadata[FLAMBE_DIRECTORIES_KEY].add(prefix[:-1])
         state_dict = self._add_registered_attrs(state_dict, prefix)
         state_dict = self._state(state_dict, prefix, local_metadata)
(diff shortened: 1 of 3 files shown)


nmatthews-asapp:
NOTE: If you change lines 644 and 650 of `tests/unit/compile/test_serialization.py` to `extra=extra` and `extra=extra2` respectively, instead of `extra=None`, this test should still pass because of the bug fixes in this branch, but 5 other tests start failing inexplicably. These line changes test having `Component`->`torch.nn.Module` relationship
@jeremyasapp @iitzco-asapp 

I think we should consider merging anyway in the interest of time

flambe/compile/component.py
```diff
@@ -976,8 +976,16 @@ def _state_dict_hook(self,
...
                                                 keep_vars=state_dict._metadata[KEEP_VARS_KEY])
                 state_dict._metadata[FLAMBE_DIRECTORIES_KEY].add(current_path)
-        # Iterate over modules to make sure Component
-        # nn.Modules are added to flambe directories
+            # Iterate over modules to make sure NON-Component
+            # nn.Modules' state is added. Only needed if self is not
+            # nn.Module, because otherwise this hook is being called
+            # via nn.Module.state_dict, and will already recurse to
+            # all children modules
+            if ((not isinstance(self, torch.nn.Module)) and isinstance(attr, torch.nn.Module)
```
iitzco-asapp:
```suggestion
            if (not isinstance(self, torch.nn.Module) and isinstance(attr, torch.nn.Module)
```

flambe/compile/component.py
```diff
@@ -976,8 +976,16 @@ def _state_dict_hook(self,
...
+                    and (not isinstance(attr, Component))):
```
iitzco-asapp: nit
```suggestion
                    and not isinstance(attr, Component):
```

flambe/compile/component.py
iitzco-asapp: What happens if `self` is a `torch.nn.Module`?

flambe/compile/serialization.py
```diff
@@ -49,7 +49,9 @@ class State(OrderedDict):
 # Private Helpers
 
 def _convert_to_tree(metadata: Dict[str, Any]) -> SaveTreeNode:
-    tree = SaveTreeNode(state={}, version=metadata[''][VERSION_KEY],
+    new_state = OrderedDict()
+    new_state._metadata = {}
```
iitzco-asapp: nit

Is it worth creating a different class for this instead of adding an attribute on the `OrderedDict`?

flambe/compile/serialization.py
```diff
@@ -97,12 +101,27 @@ def _update_save_tree(save_tree: SaveTreeNode, key: Sequence[str], value: Any) -
     last_i = 0
     for i, _ in enumerate(key):
         current_key = STATE_DICT_DELIMETER.join(key[last_i:i + 1])
+        # current_state_key = key[i]
```
iitzco-asapp: lets remove commented code.

flambe/compile/serialization.py
```diff
@@ -97,12 +101,27 @@ def _update_save_tree(save_tree: SaveTreeNode, key: Sequence[str], value: Any) -
...
+            current_key = STATE_DICT_DELIMETER.join(key[last_i:i + 1])
     current.state[current_key] = value
 
 
+def _update_save_tree_metadata(save_tree: SaveTreeNode, key: Sequence[str], value: Any) -> None:
+    current = save_tree
+    last_i = 0
+    for i, _ in enumerate(key):
+        current_key = STATE_DICT_DELIMETER.join(key[last_i:i + 1])
+        # current_state_key = key[i]
```
iitzco-asapp: Also here.

flambe/compile/serialization.py
```diff
@@ -97,12 +101,27 @@ def _update_save_tree(save_tree: SaveTreeNode, key: Sequence[str], value: Any) -
...
+        if current_key in current.children:
+            current = current.children[current_key]  # type: ignore
+            last_i = i+1
+            current_key = STATE_DICT_DELIMETER.join(key[last_i:i + 1])
```
iitzco-asapp: This has a lot of duplicated code with `_update_save_tree`. Can we extract this logic in a single method?

tests/unit/compile/test_serialization.py
```diff
@@ -107,11 +108,18 @@ def __init__(self):
 
 class BasicStateful(Component):
     def __init__(self):
-        self.x = 2019
+        self.x = random.randint(0, 100000)
```
iitzco-asapp: lets not have flaky tests here. Just use a hardcoded random number.

tests/unit/compile/test_serialization.py
```diff
@@ -107,11 +108,18 @@ def __init__(self):
...
         self.register_attrs('x')
         # self.b = Basic()
 
 
+class BasicStatefulTwo(Component, torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.y = random.randint(0, 100000)
```
iitzco-asapp: Same here, lets avoid flaky tests.

flambe/compile/component.py
nmatthews-asapp: Then torch handles the recursion

flambe/compile/serialization.py
nmatthews-asapp: This is done to be compatible with torch

tests/unit/compile/test_serialization.py
nmatthews-asapp: The number needs to be different for each initialization. We could instead have the number passed in and use an incremented variable defined in each individual test case.

iitzco-asapp: Whatever it takes to avoid `random.randint` !!

flambe/compile/serialization.py
```diff
@@ -393,7 +414,11 @@ def load_state_from_file(path: str,
                         local_metadata[FLAMBE_STASH_KEY] = stash
                     full_prefix = prefix + STATE_DICT_DELIMETER if prefix != '' else prefix
                     _prefix_keys(component_state, full_prefix)
+                    _prefix_keys(component_state._metadata, full_prefix)
```
msclar-asapp: I had issues loading this model http://10.150.0.47:49558/ due to an incompatibility issue. I fixed it by only using the _metadata attr if it exists.

```
                    if hasattr(component_state, '_metadata'):
                        _prefix_keys(component_state._metadata, full_prefix)
                        state.update(component_state)
                        # Load torch.nn.Module metadata
                        state._metadata.update(component_state._metadata)
```

Can we add this check to preserve retrocompatibility? This model was trained last week and it would become incompatible.

If not, at least can we provide an informative error message? Thank you :)

msclar-asapp: Ok, this actually didn't fix the whole issue. Still debugging, since retraining would take at least 2 days.

tests/unit/compile/test_serialization.py
```diff
@@ -106,36 +106,50 @@ def __init__(self):
 
 
 class BasicStateful(Component):
-    def __init__(self):
-        self.x = 2019
+    def __init__(self, x):
+        self.x = x
         self.register_attrs('x')
         # self.b = Basic()
```
iitzco-asapp: lets remove this.

tests/unit/compile/test_serialization.py
```diff
@@ -106,36 +106,50 @@ def __init__(self):
...
         self.linear = torch.nn.Linear(2, 2)
 
 
 class RootTorch(Component):
-    def __init__(self):
+    def __init__(self, x):
         super().__init__()
-        self.model = IntermediateStatefulTorch()
+        self.model = IntermediateStatefulTorch(x=x)
         # self.linear = torch.nn.Linear(2, 2)
```
iitzco-asapp: Same here, lets remove commented code.

nmatthews-asapp:
> Awesome fix and awesome test suite to test different combination of torch modules!
> 
> Just one last question, why `x` can't be a constant in all tests?

In some cases we're trying to test that the loaded state is in fact loaded correctly. If you initialize with exactly the same values and there is no other state, you won't be able to test this.

PR #216, opened 2020-01-23
Update requirements.txt

iitzco-asapp:
Remove `ninja` as its not being imported.


jeremyasapp:
Requirement to SRU. pinned here for some historical reason.

nmatthews-asapp:
My understanding is this is an optional requirement for SRU so we added it as a requirement to take advantage of the ninja-required features in SRU.

jeremyasapp:
Closing.

PR #224, opened 2020-01-29
Add experiment flag to pickle only on checkpoints

### LINKED ISSUE #223: Allow pickle for saving checkpoints in Experiment
Sometimes our custom save format doesn't work well for certain architectures, or users would just prefer pickle, or because the save format isn't as mature as pickle there can be bugs. For all these reasons, we should enable pickle checkpointing in Experiment

nmatthews-asapp:
Resolves #223


jeremyasapp:
I would recommend adding logic for supporting torch as the protocol

flambe/experiment/tune_adapter.py
```diff
@@ -153,13 +155,19 @@ def _train(self) -> Dict:
     def _save(self, checkpoint_dir: str) -> str:
         """Subclasses should override this to implement save()."""
         path = os.path.join(checkpoint_dir, "checkpoint.flambe")
-        self.block.save(path, pickle_only=self.pickle_checkpoints, overwrite=True)
+        if self.pickle_checkpoints:
+            torch.save(self.block, path, pickle_module=dill)
```
jeremyasapp: no overrite here?

Also still no way of selecting torch instead of dill?

PR #225, opened 2020-01-30
Add flag to TextField to save all embeddings

brianasapp:


flambe/field/text.py
```diff
@@ -85,6 +86,9 @@ def __init__(self,  # nosec
             Whether to drop tokens that don't have embeddings
             associated. Defaults to True.
             Important: this flag will only work when using embeddings.
+        setup_all_embeddings: bool
```
jeremyasapp: I don't think it's super clear what happens from this description. Maybe say that all words from the optional provided embeddings will be added to the vocabulary and to the embedding matrix

flambe/field/text.py
```diff
@@ -250,3 +238,41 @@ def process(self, example: str) -> torch.Tensor:  # type: ignore
             else:
                 ret = ret[:self.max_seq_len]
         return ret
+
+    def _get_embeddings(
+        self,
+    ) -> Tuple[KeyedVectors, List[torch.tensor]]:
+        """
```
jeremyasapp: docstring format is wrong here

brianasapp: Ah will match repo docstrings

flambe/field/text.py
brianasapp: Good point - will use that example

flambe/field/text.py
```diff
@@ -85,6 +87,10 @@ def __init__(self,  # nosec
             Whether to drop tokens that don't have embeddings
             associated. Defaults to True.
             Important: this flag will only work when using embeddings.
```
iitzco-asapp: (Out of scope but lets try to add it)

Guys wee are missing the `truncate_end` doc. @jeremyasapp can you provide a quick reference on what it does so that we can include?

flambe/field/text.py
```diff
@@ -250,3 +239,45 @@ def process(self, example: str) -> torch.Tensor:  # type: ignore
             else:
                 ret = ret[:self.max_seq_len]
         return ret
+
+    def _get_embeddings(
```
iitzco-asapp: @brianasapp this doesn't need to be an instance method. I think it would be better if we add it as a module method (passing the appropriate methods). Splitting this unnecessary coupling will make testing easier (avoiding patching for example)

iitzco-asapp: And with the benefit of having this method available for other classes/methods in the codebase.

tests/unit/field/test_text_field.py
```diff
@@ -134,6 +136,60 @@ def test_build_vocab_empty():
...
+        'green': torch.rand(5),
+        'yellow': torch.rand(5),
+        'purple': torch.rand(5),
+        'gold': torch.rand(5),
+    }
+    embedding_model.__getitem__.side_effect = embedding_model.vocab.__getitem__
+    embedding_model.__contains__.side_effect = embedding_model.vocab.__contains__
+    field._get_embeddings = MagicMock(
+        return_value=(embedding_model, [torch.rand(5)]),
+    )
```
iitzco-asapp: I don't think we should mock this method. The embeddings matrix can be provided in the constructor and we can deal with this test without using `_get_embeeddings` method right?

tests/unit/field/test_text_field.py
```diff
@@ -134,6 +136,60 @@ def test_build_vocab_empty():
...
+        setup_all_embeddings=True,
+    )
+
+    embedding_model = MagicMock()
+    embedding_model.vocab = {
+        'blue': torch.rand(5),
+        'green': torch.rand(5),
+        'yellow': torch.rand(5),
+        'purple': torch.rand(5),
+        'gold': torch.rand(5),
```
iitzco-asapp: Also, when building the embeddings, lets fix the values to avoid flakey tests in the future.

flambe/field/text.py
jeremyasapp: I didn't add that, not sure what it does.

flambe/field/text.py
brianasapp: Sure I agree, will separate it out

tests/unit/field/test_text_field.py
brianasapp: Sure but I'd like to keep the API simpler if possible

tests/unit/field/test_text_field.py
brianasapp: How could the tests be flakey

tests/unit/field/test_text_field.py
brianasapp: will still make the suggested change though

flambe/field/text.py
```diff
@@ -250,3 +266,38 @@ def process(self, example: str) -> torch.Tensor:  # type: ignore
...
+    embeddings_format: str,
+    embeddings: str,
+    embeddings_binary: bool = False,
+) -> Tuple[KeyedVectors, List[torch.Tensor]]:
+    """
+    Get the embeddings model and matrix used in the setup function
+
+    Returns
+    -------
+    KeyedVectors
```
iitzco-asapp: Return type doesn't match signature

iitzco-asapp: Actually I think you want to keep `KeyedVectors`!

PR #233, opened 2020-02-07
(Not so) Big refactor

iitzco-asapp:


flambe/field/text.py
```diff
@@ -274,6 +374,10 @@ def with_embeddings(
         embeddings_binary : bool, optional
             Whether the input embeddings are provided in binary format,
             by default False
+        setup_all_embeddings: bool
```
brianasapp: Will this also be removed from the constructor?

flambe/field/text.py
```diff
@@ -138,62 +199,100 @@ def vocab_size(self) -> int:
...
+                        new_vocab[token] = new_index = new_index + 1
                     else:
-                        self.vocab[token] = index = index + 1
+                        # Collapse all OOV's to the same <unk> token id
+                        new_vocab[token] = new_vocab[self.unk]
+
+        if self.setup_all_embeddings:
+            for token in model.vocab.keys():
+                if token not in new_vocab:
+                    new_vocab[token] = index
```
brianasapp: nit: this index setter is written differently than the rest of the file

brianasapp: I think this should be `new_index`

flambe/field/text.py
```diff
@@ -97,6 +149,19 @@ def __init__(self,  # nosec
             embedding matrix. Defaults to False.
 
         """
+        if embeddings:
+            if model:
+                raise ValueError("Cannot submit a model and use the embeddings parameters" +
+                                 "simultaneously. Use the 'from_embeddings' factory instead.")
+
+            warnings.warn("The embeddings parameters will be deprecated in future release. " +
```
brianasapp: Should the warning include `setup_all_embeddings` or is it obvious?

flambe/field/text.py
```diff
@@ -138,62 +199,100 @@ def vocab_size(self) -> int:
...
+        -------
+        Tuple[OrderedDict, torch.Tensor]
+            A tuple with the new embeddings and the embedding matrix
+        """
+        embedding_matrix: List[torch.Tensor] = []
+        new_vocab: odict[str, int] = odict()
+
+        new_index = -1
+
+        for token, index in self.vocab.items():
```
brianasapp: Can this be a single loop since the `index` value is unused?

Suggestion:
```
tokens = self.vocab.values()
if self.setup_all_embeddings:
    tokens.update(model.vocab.keys())

for token in tokens:
    ...

File: flambe/field/text.py (hunk 1 above)
Comment (iitzco-asapp) [reply flambe/field/text.py]: Yes, as this param only works when there are embeddings

File: flambe/field/text.py (hunk 2 above)
Comment (iitzco-asapp) [reply flambe/field/text.py]: good catch. Done.

File: flambe/field/text.py [hunk 5]
```diff
@@ -138,62 +199,100 @@ def vocab_size(self) -> int:
...
+            if token in self.specials:
```
brianasapp: one more suggestion:

this can be a single if else statement I think:

```
if token in model:
    ...
elif token in self.specials:
    emb.append(torch.randn(...))
    new_vocab[token] = ...
elif self.unk_init_all:
    ...
else:
    ...
```

Might be clearer and can get rid of the ternary for the specials

flambe/field/text.py
iitzco-asapp: Not only `setup_all_embeddings` but also other embeddings-specific parameters will go away. Adding this to the warning.

flambe/field/text.py
iitzco-asapp: Sorry missed this!

iitzco-asapp: Yes, excellent.

PR #234, opened 2020-02-07
Create publishing file

jeremyasapp:
Github action for publishing Flambé


iitzco-asapp:
I would suggest first publishing to test.pypi before pushing to the official pypi.

iitzco-asapp:
It is a useful security check to avoid messing releases up

PR #242, opened 2020-02-11
AUC metric returns number instead of tensor

laberastury-asapp:


flambe/metric/dev/auc.py
```diff
@@ -64,7 +64,7 @@ def finalize(self, state: Dict) -> float:
             return np.NaN
         pred = torch.cat(state['pred'], dim=0)
         target = torch.cat(state['target'], dim=0)
-        state['accumulated_score'] = self.compute(pred, target)
+        state['accumulated_score'] = self.compute(pred, target).item()
```
iitzco-asapp: its awkward that this method modifies the state of the param right? Why is this happening?

laberastury-asapp: Good point, will investigate to see why this is so.

cle-ros: because it is modified in-place. that makes keeping track of the states in the train class a lot easier

laberastury-asapp: Yes, looks like all metrics are using this state dict to setup these predefined properties (`accumulated_score` is adjusted in `aggregate()`). This pattern where we set in the state and also return from it is only in `AUC`, and changing it would break the consistency with the other metrics. We could do a refactor to make this clearer (e.g. the metric could own its state, instead of an external dictionary), but since it wouldn't be a small change I think it's out of scope for this PR.

jeremyasapp: I don't like the metric being stateful, what's wrong with passing the state around?

laberastury-asapp: The thing is that it does have state, it's just represented outside the object in the dictionary. This is problematic because it's leaked out (+ the drawbacks of using a dictionary to represent behavior in string keys). Anyway indeed it might not be needed in Metric (as per the class that does calculation), maybe it's more of a runtime metric wrapper over metric + state.

jeremyasapp: yes but the state doesn't need to live inside the object, that s my point

PR #247, opened 2020-02-17
Create abstraction for embeddings to avoid using KeyedVectors

iitzco-asapp:
Created `PretrainedEmbeddings` as an abstraction for pretrained embeddings.

This avoids the dependency on KeyedVectors

This change is fully compatible with `0.4.12` and `0.4.13`


jeremyasapp:
Closing for now, as not really needed

PR #248, opened 2020-02-19
Lazy TextField

iitzco-asapp:
Because of how serialization works, the `TextField` will be initialized each time it is de-serialized.

Because of this behavior, we don't want the embedding model to be loaded in the init, mainly because the embedding model is not used when `setup` is not needed (this is the case for inference models)

This PR initializes the embedding model in the `setup` method rather than in the init, keeping object initialization super lightweight.


flambe/field/text.py
```diff
@@ -65,6 +65,46 @@ def get_embeddings(
...
+        Whether to drop tokens that don't have embeddings
+        associated. Defaults to True.
+        Important: this flag will only work when using embeddings.
+
+    """
+    embeddings: str
+    embeddings_format: str = 'gensim'
+    embeddings_binary: bool = False
+    setup_all_embeddings: bool = False
+    unk_init_all: bool = False
```
jeremyasapp: this is not necessary embedding related. Neither is drop_unknown

flambe/field/text.py
```diff
@@ -90,10 +130,10 @@ def __init__(self,  # nosec
                  unk_token: str = '<unk>',
                  sos_token: Optional[str] = None,
                  eos_token: Optional[str] = None,
+                 embeddings_info: Optional[EmbeddingsInformation] = None,
```
jeremyasapp: Why did you add this? Seems unnecessary, and adds an extra level of nesting

flambe/field/text.py
iitzco-asapp: iirrc and according to the documentation, this parameters are only used in case embeddings are provided. Do you have an example where they are not used with embeddings?

iitzco-asapp: `unk_init_all` is for for the vocab that is not in the embeddings.

jeremyasapp: well for drop_unkown you can drop things that are not in the vocab, and you may have created the vocab from the data not the embeddings file

flambe/field/text.py
```diff
@@ -65,6 +65,46 @@ def get_embeddings(
...
+    drop_unknown: bool
+        Whether to drop tokens that don't have embeddings
+        associated. Defaults to True.
+        Important: this flag will only work when using embeddings.
+
+    """
+    embeddings: str
+    embeddings_format: str = 'gensim'
+    embeddings_binary: bool = False
+    setup_all_embeddings: bool = False
```
jeremyasapp: I don't like this parameter name, it's not clear that it has to do with using the embeddings as source of the vocabulary

flambe/field/text.py
iitzco-asapp: and `drop_unknown` just works now when embeddings are passed. Maybe we can generalize this one to work also for non embedding TextFields

flambe/field/text.py
iitzco-asapp: By having all related parameters in the same data structure, the code becomes:

* more clear and readable, following a composition pattern allowing the creation of factories 
* easier to test
* easier to extend the class as your __init__ doesn't have tons of parameters anymore
* easier to implement custom logic in this new class rather than in the TextField iteself. For example, avoiding having in the TextField checks like "user passed the `embedding` name but `embedding_format` is None" which can be delegated to the appropriate class.

Users won't have another level of nesting in they configs as they will use the factory

flambe/field/text.py
iitzco-asapp: > well for drop_unkown you can drop things that are not in the vocab, and you may have created the vocab from the data not the embeddings file

Yes lets generalize this concept

flambe/field/text.py
iitzco-asapp: `embeddings_as_vocab_source`? Better ideas?

brianasapp: `use_all_from_vectored_keys`?

flambe/field/text.py
```diff
@@ -258,7 +325,7 @@ def _build_embeddings(self, model: KeyedVectors) -> Tuple[odict, torch.Tensor]:
 
         tokens: Iterable[str] = self.vocab.keys()
 
-        if self.setup_all_embeddings:
+        if setup_all_embeddings:
```
axel-asapp: I would rename to avoid the outer scope naming collapse. It would become clearer it is definitely a local variable

flambe/field/text.py
```diff
@@ -272,7 +339,7 @@ def _build_embeddings(self, model: KeyedVectors) -> Tuple[odict, torch.Tensor]:
                 else:
                     self.unk_numericals.add(self.vocab[token])
 
-                    if self.unk_init_all:
+                    if unk_init_all:
```
axel-asapp: I would rename to avoid the outer scope naming collapse. It would become clearer it is definitely a local variable

flambe/field/text.py
axel-asapp: I think is a great idea to take subclassing and logic extraction into appropiate components 👏

flambe/field/text.py
iitzco-asapp: Lets do `build_vocab_from_embeddings`.

Already discussed with the team, this is a harmless backwards incompatible change.

iitzco-asapp: Actually, no backwards incompatibility as we keep `setup_all_embeddings` in the init

PR #251, opened 2020-02-21
Change comparison when loading from pickle to not break when comparing tensors

msclar-asapp:
When loading tensors from pickle, there's a comparison to trigger a warning if PickledDataLink was called twice with different stashes.

This comparison breaks when using tensors, because == comparison returns a tensor. 

For example, 

```
if torch.Tensor([False, False]): 
    pass
```

Will return `RuntimeError: bool value of Tensor with more than one value is ambiguous`.

Now, there's a distinction being made to check that all elements in a tensor are exactly the same. I wasn't able to fix this without making this explicit distinction, but please let me know if you know how!

### DIFF
--- a/flambe/compile/component.py
+++ b/flambe/compile/component.py
@@ -368,7 +368,11 @@ class PickledDataLink(Registrable):
 
     def __call__(self, stash: Dict[str, Any]) -> Any:
         if self.obj_value is not None:
-            if self.obj_value != stash[self.obj_id]:
+            diff_obj_stash = self.obj_value != stash[self.obj_id]
+            is_tensor = type(diff_obj_stash) == torch.Tensor
+
+            # == comparison between tensors returns a tensor, not bool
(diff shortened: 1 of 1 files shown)


flambe/compile/component.py
```diff
@@ -368,7 +368,11 @@ def __init__(self, obj_id: str, value: Any = None):
 
     def __call__(self, stash: Dict[str, Any]) -> Any:
         if self.obj_value is not None:
-            if self.obj_value != stash[self.obj_id]:
+            diff_obj_stash = self.obj_value != stash[self.obj_id]
+            is_tensor = type(diff_obj_stash) == torch.Tensor
```
nmatthews-asapp: I would recommend using `isinstance` rather than equality on the type in general unless you have a specific reason for not doing that here?

msclar-asapp: It makes total sense! That way we also include subclasses. Changed!

PR #254, opened 2020-02-21
Added BaseSampler "downsample_max_samples" constructor argument

cle-ros:
changed basesampler to take an additional 'downsample_max_samples' in the constructor that behaves exactly like 'downsample', but does not specify a ratio but an absolute number of samples instead

### DIFF
--- a/flambe/sampler/base.py
+++ b/flambe/sampler/base.py
@@ -1,4 +1,5 @@
 import math
+import warnings
 from collections import defaultdict, OrderedDict as odict
 from itertools import chain
 from functools import partial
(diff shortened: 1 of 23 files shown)


flambe/sampler/base.py
```diff
@@ -264,15 +274,24 @@ def sample(self,
         if len(data) == 0:
             raise ValueError("No examples provided")
 
-        if self.downsample:
-            if not (0 < self.downsample <= 1):
+        if self.downsample or self.downsample_max_samples:
```
jeremyasapp: we may want to use is not None here in case people use 0, even if that s weird

cle-ros: done

PR #255, opened 2020-02-21
Text- and TransformerField now take nested objects (dicts and lists) as arguments to process and setup

cle-ros:
changed the text and transformer fields to also take a (nested) list of strings of dict of strings to both setup and process.

required for compatibility with specific datasets


flambe/field/text.py
```diff
@@ -295,7 +319,8 @@ def setup(self, *data: np.ndarray) -> None:
             self.vocab, self.embedding_matrix = self._build_embeddings(self.model)
 
     # TODO update when we add generics
-    def process(self, example: str) -> torch.Tensor:  # type: ignore
+    def process(self, example: Union[str, List[Any], Dict[Any, Any]]) \
```
jeremyasapp: docstring isn't quite right. Should be  `Union[str, List[str], Dict[Any, str]]`

cle-ros: What if you have nested objects? E.g., `{'asdf': ['asdf', 'asdf'], {'asdf': {'asdf': ['asdf', 'asdf'], 'a2': 'asdf'}}}`

jeremyasapp: Does the PR support that?

cle-ros: yes, completely. There are even testcases for this.

PR #257, opened 2020-02-21
Can specify more extra tokens that should be included in vocab in 'from_embeddings'

cle-ros:
- added the ability to provide arbitrary special tokens to the from_embeddings method of textfield

- added unit tests for this case

### DIFF
--- a/tests/unit/field/test_text_field.py
+++ b/tests/unit/field/test_text_field.py
@@ -254,6 +254,7 @@ def test_load_embeddings_with_extra_tokens():
     assert '<a>' in field.vocab and '<b>' in field.vocab
     assert '<c>' not in field.vocab
     assert field.embedding_matrix[field.vocab['<a>']].size(-1) == 4
+    assert field.embedding_matrix[field.vocab['<b>']].size(-1) == 4
 
 
 def test_load_embeddings_legacy():
@@ -360,12 +361,12 @@ def recursive_tensor_to_list(data):
     if isinstance(data, list):
         return [recursive_tensor_to_list(d) for d in data]
     elif isinstance(data, dict):
-        return dict((k, recursive_tensor_to_list(v)) for k, v in data.items())
+        return {(k, recursive_tensor_to_list(v)) for k, v in data.items()}
     elif isinstance(data, torch.Tensor):
         return data.tolist()
 
 
-def test_load_embeddings_with_extra_tokens():
+def test_setup_with_extra_tokens():
     field = TextField.from_embeddings(
         embeddings="tests/data/dummy_embeddings/test.txt",
         pad_token=None,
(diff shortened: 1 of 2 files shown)


tests/unit/field/test_text_field.py
```diff
@@ -242,6 +242,20 @@ def test_load_embeddings():
...
+        embeddings="tests/data/dummy_embeddings/test.txt",
+        pad_token=None,
+        unk_init_all=False,
+        additional_vocab_tokens=['<a>', '<b>', '<c>']
+    )
+    dummy = "a test ! <a> <b> "
+    field.setup([dummy])
+    assert '<a>' in field.vocab and '<b>' in field.vocab
+    assert '<c>' not in field.vocab
+    assert field.embedding_matrix[field.vocab['<a>']].size(-1) == 4
```
iitzco-asapp: assert than `<b>` is also part of the embedding_matrix?

tests/unit/field/test_text_field.py
```diff
@@ -340,3 +354,31 @@ def test_text_process_unk():
     dummy = "justo Praesent luctus justo praesent"
     with pytest.raises(Exception):
         field.process(dummy)
+
+
+def recursive_tensor_to_list(data):
+    if isinstance(data, list):
+        return [recursive_tensor_to_list(d) for d in data]
+    elif isinstance(data, dict):
+        return dict((k, recursive_tensor_to_list(v)) for k, v in data.items())
```
iitzco-asapp:
```suggestion
        return {k: recursive_tensor_to_list(v)) for k, v in data.items()}
```

tests/unit/field/test_text_field.py
```diff
@@ -340,3 +354,31 @@ def test_text_process_unk():
...
+        unk_init_all=False,
+        additional_vocab_tokens=['<a>', '<b>', '<c>']
+    )
+
+    dummy = "this is a test"
+    field.setup([dummy])
+    assert recursive_tensor_to_list(field.process(dummy)) == [1, 2, 3, 4]
+
+    dummy = "this is a test <a>"
+    assert recursive_tensor_to_list(field.process(dummy)) == [1, 2, 3, 4, 0]
```
iitzco-asapp: nit
For now this is fine, but its worth thinking how to decouple the implementation logic from the assertion (we are assuming that the `additional_vocab_tokens` are inserted in the vocab before the data tokens)

For example, we could do stuff like this
```
assert recursive_tensor_to_list(field.process(dummy))[-1] == field.vocab['<a>']
```

cle-ros: I agree. But that would be a general change to most of the different TextField tests.

tests/unit/field/test_text_field.py
cle-ros: done

tests/unit/field/test_text_field.py
cle-ros: done

cle-ros:
replaced by https://github.com/asappresearch/flambe/pull/274

PR #262, opened 2020-02-24
Perplexity Hotfix

cle-ros:
Hotfix for perplexity, because it requires special aggregation behavior

### DIFF
--- a/flambe/metric/dev/perplexity.py
+++ b/flambe/metric/dev/perplexity.py
@@ -1,3 +1,5 @@
+from typing import Dict
+import numpy as np
 import torch
 
 from flambe.metric import Metric
@@ -8,7 +10,7 @@ class Perplexity(Metric):
 
     def __init__(self):
         """Perplexity, computed as CrossEntropy"""
-        self.entropy = torch.nn.CrossEntropyLoss()
+        self.entropy = torch.nn.CrossEntropyLoss(reduction='none')
 
     def compute(self, pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         """Compute the preplexity given the input and target.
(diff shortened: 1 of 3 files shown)


flambe/metric/dev/perplexity.py
```diff
@@ -28,3 +30,42 @@ def compute(self, pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
...
+
+        Returns
+        -------
+        dict
+            the state dict
+        """
+        pred, target = args
+        if not state:
+            state['historic_score'] = []
+        state['historic_score'].append(self.entropy(pred, target).cpu().detach())
```
jeremyasapp: I think we can be more efficient by keeping only the sum:
```python
logits = self.entropy(pred, target).cpu().detach()
state['score'] += logits.sum()
state['num_samples'] += len(logits)
```
Then in finalize:
```python
logits = self.entropy(pred, target).cpu().detach()
return torch.exp(state['scores'] / state['num_samples']).item()
```

flambe/metric/dev/perplexity.py
```diff
@@ -28,3 +30,42 @@ def compute(self, pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
...
+
+        Parameters
+        ----------
+        state: dict
+            the metric state
+
+        Returns
+        -------
+        float
+            The final score.
```
jeremyasapp: make sure to leave a blank line at the end of the docstring

flambe/metric/dev/perplexity.py
cle-ros: done

flambe/metric/dev/perplexity.py
cle-ros: done

cle-ros:
all done;
also included more test cases

PR #263, opened 2020-02-24
Add mult-gpu support to train.py

schhabria-asapp:


jeremyasapp:
There are some challenges I hadn't thought of with the current Trainer implementation, since 0.5 will be released this week, I suggest we wait for that.

PR #269, opened 2020-02-26
github action clean up

yukw777:
Run only the fast tests (unit and end2end for now) on every push to a branch that is not master.
Run only the fast tests (unit and end2end for now) on every pull request.
Run all the tests on push to master.

### DIFF
--- a/.github/workflows/tests-all.yml
+++ b/.github/workflows/tests-all.yml
@@ -1,14 +1,13 @@
-name: Run slow tests
+name: Run all tests
 
 on:
-  schedule:
-    - cron: "0 0 * * *"
+  # run all the tests and sonar cloud on push to master
   push:
     branches:
       - master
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tests-all.yml
```diff
@@ -35,11 +34,16 @@ jobs:
         pip install --upgrade -r tests/requirements.txt
         pip install tox awscli
         python -m awscli configure set region us-east-1
-    - name: Text examples
+    - name: Run all the tests
       run: |
-        python3 -m tox -- -v -m "examples"
+        tox -- -v
```
iitzco-asapp: This will not allow parallelizing these tests. And it may not be safe for coverage.

yukw777:
closing in favor of https://github.com/asappresearch/flambe/pull/270

PR #279, opened 2020-03-20
Allow parallel runs on the same machine with multiple GPUs

jeremyasapp:


flambe/experiment/experiment.py
```diff
@@ -274,8 +274,12 @@ def run(self, force: bool = False, verbose: bool = False, debug: bool = False, *
 
         # By default use all CPUs if no GPU is present
         devices = self.devices if self.devices else None
-        if devices is None and utils.local_has_gpu():
```
nmatthews-asapp: I believe this is the only usage of this util so it should be deleted right?

flambe/experiment/experiment.py
```diff
@@ -274,8 +274,12 @@ def run(self, force: bool = False, verbose: bool = False, debug: bool = False, *
...
-            devices = {"cpu": 4, "gpu": 1}
+        if devices is None:
+            cluster_devices = ray.cluster_resources()
+            if 'GPU' in cluster_devices or 'gpu' in cluster_devices:
+                devices = {"cpu": 4, "gpu": 1}
```
nmatthews-asapp: Does this cover all cases? i.e. do we not support parallelizing variants on a single multi-gpu machine?

jeremyasapp: we do, this is exactly what this PR addresses. This is only for when devices is not provided as a sensible default, otherwise the user should specify

flambe/experiment/experiment.py
jeremyasapp: Yeah I think that's true, but I figured we might want to reuse that some day. Either way, it's gone in 0.5

flambe/experiment/experiment.py
```diff
@@ -274,8 +274,12 @@ def run(self, force: bool = False, verbose: bool = False, debug: bool = False, *
 
         # By default use all CPUs if no GPU is present
         devices = self.devices if self.devices else None
-        if devices is None and utils.local_has_gpu():
-            devices = {"cpu": 4, "gpu": 1}
+        if devices is None:
+            cluster_devices = ray.cluster_resources()
+            if 'GPU' in cluster_devices or 'gpu' in cluster_devices:
```
iitzco-asapp: nit for avoiding repeating the check:
```
if any(x in cluster_devices for x in ['GPU', 'gpu']):
```
or
```
if any(map(lambda x: x in cluster_devices, ['GPU', 'gpu'])):
```

flambe/experiment/experiment.py
```diff
@@ -274,8 +274,12 @@ def run(self, force: bool = False, verbose: bool = False, debug: bool = False, *
...
+                devices = {"cpu": 4, "gpu": 1}
+            else:
+                devices = {"cpu": 1}
```
iitzco-asapp: I would extract these 4 lines into a utility method and add unit tests, as its critical to maintain the correct functionality.

brendandc-asapp: :+1: on adding tests!

jeremyasapp: Since this code is completely going away in the next version, can I get a pass here? This fixes a big bug, and this code is very minimal

flambe/experiment/experiment.py
jeremyasapp: I actually find that harder to read TBH. I would do it if there were more than 2, but here I feel like it's okay

PR #280, opened 2020-03-25
Add requirements file to the output folder

iitzco-asapp:
When saving an artifact, a `requirements.txt` file will be included with the frozen dependencies used to produce that artifact.

This feature will not be available when using the `pickle` option.


tests/unit/compile/test_utils.py
```diff
@@ -0,0 +1,12 @@
+import tempfile
+
+
+from flambe.compile.utils import write_deps
+
+
+def test_write_deps():
+    dummy_dependencies = ['numpy==1.2.3', 'pip~=1.1.1', 'some_other-random dep']
+    with tempfile.NamedTemporaryFile() as tmpfile:
+        write_deps(tmpfile.name, dummy_dependencies)
```
brendandc-asapp: i think you should test it without the input as well - preferably with a mock on freeze

iitzco-asapp: yes good idea.

PR #284, opened 2020-03-30
Hotfix ray debug jeremyasapp

jeremyasapp:
- Move the default CPU and GPU counts to a utility function
- Add tests
- Fixes a bug in debug mode


flambe/experiment/utils.py
```diff
@@ -549,3 +549,39 @@ def shutdown_remote_ray_node(host: str,
...
+
+def get_default_devices(debug: bool = False,
+                        default_cpus: int = 1,
+                        default_gpus: int = 1) -> Dict[str, int]:
+    """Get the default devices to use if not provided.
+
+    Parameters
+    ----------
+    debug: bool, optional
+        If we are running in debug mode (where Ray is not available)
```
iitzco-asapp:
```suggestion
        If we are running in debug mode (where Ray is not available).
        If debug is False, this method should be called after running
        ``ray.init()``.
```

flambe/experiment/utils.py
```diff
@@ -549,3 +550,47 @@ def shutdown_remote_ray_node(host: str,
...
+    if debug:
+        if torch.cuda.is_available():
+            devices = {"cpu": default_cpus, "gpu": default_gpus}
+        else:
+            devices = {"cpu": default_cpus}
+    else:
+        if ray.is_initialized():
+            cluster_devices = ray.cluster_resources()
+            use_gpu = 'GPU' in cluster_devices or 'gpu' in cluster_devices
+        elif torch.cuda.is_available():
```
iitzco-asapp: I would just fail if ray wan't initialized

iitzco-asapp: to not make the logic more complex

jeremyasapp: disagree, this function is more powerful as is!

tests/unit/experiment/test_utils.py
```diff
@@ -112,3 +113,26 @@ def test_divide_nested_grid_search_options_nested_options(make_classes):
...
     divided_configs = list(divide_nested_grid_search_options(config))
     assert repr(divided_configs) == repr([config1, config2])
+
+
+def test_default_devices():
+    ray.init()
+    devices = get_default_devices(debug=False)
+    assert devices['cpu'] == 1
+
+    devices = get_default_devices(debug, default_cpus=2)
```
iitzco-asapp: `debug` has no value here

iitzco-asapp: forgot `debug=False`?

iitzco-asapp: I think you forgot `debug=False` as the first line of the test

tests/unit/experiment/test_utils.py
```diff
@@ -112,3 +113,26 @@ def test_divide_nested_grid_search_options_nested_options(make_classes):
...
+
+
+def test_default_devices_debug():
+    devices = get_default_devices(debug=True)
+    assert devices['cpu'] == 1
+
+    devices = get_default_devices(debug, default_cpus=2)
+    assert devices['cpu'] == 2
+
+    devices = get_default_devices(debug, default_gpus=2)
```
iitzco-asapp: Same here

tests/unit/experiment/test_utils.py
jeremyasapp: yep my bad

tests/unit/experiment/test_utils.py
jeremyasapp: fixed

jeremyasapp:
> Testing will be more robust mocking both `torch.cuda.is_available` and `ray.cluster_resources()`.
> 
> That's how you can test all possible flows.

Is that even possible? I think this is probably fine

iitzco-asapp:
> Is that even possible? I think this is probably fine

Sure, you can mock anything you want. For example:

```
@mock.patch("flambe.experiment.utils.torch.is_cuda_available")
def test_case(mock_cuda_available):
     mock_cuda_available.return_value = False
```

jeremyasapp:
Added mocks to have full coverage

PR #289, opened 2020-05-13
Sample new embedding from maximum likelihood estimate

jeremyasapp:
This PR improves over the status quo of producing random embeddings for unknown tokens by learning a normal distribution over the provided embeddings.


jeremyasapp:
The sonar error is due to the fact that i had to fix some flake8 errors across the code base. Should be ignored

